PR #3, opened 2021-02-10
Fix github container registry

maruina:
It seems I was following the wrong tutorial.
- Fix the login steps as per https://github.com/docker/login-action#github-container-registry
- ~Created a PAT as per https://docs.github.com/en/packages/guides/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry~
- Push action changed as per https://github.com/docker/build-push-action#push-to-multi-registries

### DIFF
--- a/.github/workflows/release-artifacts.yaml
+++ b/.github/workflows/release-artifacts.yaml
@@ -25,12 +25,15 @@ jobs:
       uses: olegtarasov/get-tag@v2
       id: tagName
 
-    - name: Build on PR
-      id: docker_build
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v1
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.CR_PAT }}
+
+    - name: Build and publish a tagged docker image
       uses: docker/build-push-action@v2
       with:
-        username: ${{ github.actor }}
-        password: ${{ secrets.GITHUB_TOKEN }}
-        registry: docker.pkg.github.com
-        repository: ${{ github.repository }}/argocd-progressive-rollout
-        tags: ${{ steps.tagName.outputs.tag }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-artifacts.yaml
```diff
@@ -25,12 +25,15 @@ jobs:
...
       id: tagName
 
-    - name: Build on PR
-      id: docker_build
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v1
+      with:
+        registry: ghcr.io
+        username: ${{ github.repository_owner }}
+        password: ${{ secrets.CR_PAT }}
```
sledigabel, line 33: Looks like this is something we need to create manually this time, I assume you've done this already? Is this a company-shared secret?

sledigabel, line 33: we prob don't need to share the PAT anywhere, and will invalidate/regenerate if we need?

sledigabel (approved):
looks nice and clean; fixed the github package auth and push part

.github/workflows/release-artifacts.yaml
maruina: Yes, if we need a new one any owner can do it.

maruina:
Pending enabling `GitHub Container Registry` for our organization.

maruina:
Had a chat with @porelle-logan and he advised to stick to the "old" GitHub Packages Docker registry.

sledigabel (approved):
LGTM

.github/workflows/release-artifacts.yaml
maruina: And now we don't need that until we migrate to the new registry.

maruina:
And it's working :)
https://github.com/Skyscanner/argocd-progressive-rollout/packages/618978

PR #16, opened 2021-02-16
Add basic controller structure

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/7

- Improve README with motivations, example spec and link to contributing
- Add a progressive rollout object sample
- Add object status
- Add tests for manipulating object status
- Add simple test for the reconciliation loop
- Fix golangci-lint pre-commit warning about not using a pinned version
- Fix API group from `deployment.skyscanner.net.deployment.skyscanner.net` to `deployment.skyscanner.net`


config/samples/deployment.skyscanner.net_v1alpha1_progressiverollout.yaml
```diff
@@ -1,7 +1,33 @@
 apiVersion: deployment.skyscanner.net.deployment.skyscanner.net/v1alpha1
```
sledigabel, line 1: is this a test that gets run?

sledigabel (dismissed):
LGTM, adds a simple and minimal structure for the progressive rollout spec with documentation and example.

config/samples/deployment.skyscanner.net_v1alpha1_progressiverollout.yaml
maruina: I think that I wrote the domain twice when initializing the project. Good catch.

README.md
```diff
@@ -2,9 +2,57 @@
...
-`pre-alpha`. Expect breaking changes until `v0.1.0`.
+[Argo ApplicationSet](https://github.com/argoproj-labs/applicationset) is being developed as the solution to replace the `app-of-apps` pattern.
+
+While `ApplicationSet` is great to programmatically generate ArgoCD Applications, we will still need to solve _how_ to update the Applications.
+
+If we enable the `auto-sync` policy, we will update all the generated Applications at once.
+
+This might not be a problem if we have only one production cluster, but organizations with tens or hundreds of clusters need to avoid a global rollout. They need to release new versions of their application in a safer way.
+
+The `argocd progressive rollout` controller allow operators and developers to decide _how_ they want to update their Applications.
```
Smirl, line 15:
```suggestion
The `argocd progressive rollout` controller allows operators and developers to decide _how_ they want to update their Applications.
```

README.md
```diff
@@ -2,9 +2,57 @@
...
+
+```yaml
+apiVersion: deployment.skyscanner.net/v1alpha1
+kind: ProgressiveRollout
+metadata:
+  name: myservice
+  namespace: argocd
+# the rollout steps
+stages:
+  # human friendly name
```
Smirl, line 27:
```suggestion
   # human friendly name
```

api/v1alpha1/progressiverollout_types.go
```diff
@@ -28,18 +30,115 @@ type ProgressiveRolloutSpec struct {
...
+	Status metav1.ConditionStatus `json:"status"`
+	// Reason is a one-word CamelCase reason for the condition's last transition
+	Reason string `json:"reason,omitempty"`
+	// LastTransitionTime of this condition
+	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+}
+
+type ProgressiveRolloutConditionType string
 
+const CompletedType ProgressiveRolloutConditionType = "Completed"
```
sledigabel, line 92: perhaps worth adding more types, at least "Failed"?

sledigabel: I would actually name it `CompletedCondition` rather than `Type`? That's not really at type any more at this stage?

api/v1alpha1/progressiverollout_types_test.go
```diff
@@ -0,0 +1,54 @@
...
+			status.SetCondition(&empty)
+			nonEmpty := status.NewCondition(CompletedType, metav1.ConditionTrue, "Completed")
+			status.SetCondition(&nonEmpty)
+			Expect(status.Conditions[0]).To(Equal(nonEmpty))
+			Expect(len(status.Conditions)).To(Equal(1))
+		})
+
+		It("should not update the condition if it already exists and has the same status and reason.", func() {
+			status := &ProgressiveRolloutStatus{}
+			first := status.NewCondition(CompletedType, metav1.ConditionFalse, "Failed")
```
sledigabel, line 36: In tests we might want to refer to `CompletedCondition` or `FailedCondition` rather than a string?

sledigabel: unless I'm not getting what the SetCondition does

api/v1alpha1/progressiverollout_types.go
sledigabel: Let's settled for `CompletedConditionType` :-)

sledigabel:
Just a few minor comments, but as a basic structure I'm quite happy with this.
Perhaps let's make it clear in the documentation this is non functional at the moment?

maruina:
Sure, I'll update with your comments and amend the README.

api/v1alpha1/progressiverollout_types.go
maruina: So, the reason is why I have only Completed, it's because that you have the `Status` that can be `true` or `false`. If we add the Failed one, than we should set Failed to true? If we look at https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status they suggest to use a positive context, so we have a ProgressiveRollout that can be Completed=true or Completed=false. Does it make sense?

maruina: I think in the DR we called the condition `RolledOut`, so if that's more clear I'm happy to change it.

sledigabel: OK I get it now, thanks for the pointer.
In this case `Completed` actually makes more sense. I'm concerned `RolledOut` could potentially create confusion down the line with concepts like "Rolledback"?
From your link, we need to differentiate all those states (conditions) with the `name`.

Nice one

maruina:
I spent some times looking at the gitops tookit controller, in particular the [kustomize-controller](https://github.com/fluxcd/kustomize-controller).

I discovered that `k8s.io/apimachinery v0.20.1` includes a [Condition type](https://pkg.go.dev/k8s.io/apimachinery@v0.20.2/pkg/apis/meta/v1#Condition) and [some methods](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#SetStatusCondition) to interact with that.

I pushed a new code change to use the apimachinery condition and remove a lot of custom code.

api/v1alpha1/progressiverollout_types.go
```diff
@@ -28,18 +30,68 @@ type ProgressiveRolloutSpec struct {
...
+}
+
+// ProgressiveRolloutStage defines a rollout stage
+type ProgressiveRolloutStage struct {
+	// Name is a human friendly name for the stage
+	//+kubebuilder:validation:Required
+	Name string `json:"name"`
+	// MaxParallel is how many selected targets to update in parallel
+	//+kubebuilder:validation:Minimum:1
+	MaxParallel intstr.IntOrString `json:"maxParallel"`
```
sledigabel, line 48: `IntOrString` is this an option? or is that to allow things like `100%`?

api/v1alpha1/progressiverollout_types.go
```diff
@@ -28,18 +30,68 @@ type ProgressiveRolloutSpec struct {
...
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
-// +kubebuilder:object:root=true
+// GetStatusConditions returns a pointer to the Status.Conditions slice
+func (in *ProgressiveRollout) GetStatusConditions() *[]metav1.Condition {
+	return &in.Status.Conditions
+}
 
+func (in *ProgressiveRollout) NewStatusCondition(t string, s metav1.ConditionStatus, r string, m string) metav1.Condition {
```
sledigabel, line 83: Should the type be its own type and we could use constant to keep consistency across conditions?
(that is, if i understand it well, it would be something like "Deploying", "Deployed", "Failed" or "RolledBack"?

controllers/progressiverollout_controller.go
```diff
@@ -34,15 +35,26 @@ type ProgressiveRolloutReconciler struct {
...
-	// your logic here
+	// Get the ProgressiveRollout object
+	pr := deploymentskyscannernetv1alpha1.ProgressiveRollout{}
+	if err := r.Get(ctx, req.NamespacedName, &pr); err != nil {
+		log.Error(err, "unable to fetch ProgressiveRollout", "object", pr.Name)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
 
+	// Rollout completed
+	completed := pr.NewStatusCondition(deploymentskyscannernetv1alpha1.CompletedCondition, metav1.ConditionTrue, deploymentskyscannernetv1alpha1.StagesCompleteReason, "All stages completed")
```
sledigabel, line 52: For now we simply set it to completed, is this right?

api/v1alpha1/progressiverollout_types.go
maruina: That is coming from https://pkg.go.dev/k8s.io/apimachinery/pkg/util/intstr@v0.20.1. It's used as well in the deployment controller so you can specify an integer or a %, see https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go

api/v1alpha1/progressiverollout_types.go
maruina: The reason why I'm using `t string` (t as type) it's because `metav1.Condition` takes a string as type.

My understanding is that if we were using our own type, we would have a type mismatch, unless there is something I am missing.

controllers/progressiverollout_controller.go
maruina: Yeah, so at the moment we have just one type `Completed` that is `true` only when all the stages are completed.

api/v1alpha1/progressiverollout_types.go
sledigabel: As discussed, I finally get that we're going to keep it relatively light and native on the conditions, and will focus on a dedicated free form structures in `status` later to report on progress.

api/v1alpha1/progressiverollout_types.go
sledigabel: as i suspected. thank you for clarifying.

api/v1alpha1/progressiverollout_types.go
sledigabel: Gotcha. we'll keep it simple on the conditions

sledigabel (dismissed):
LGTM,
Simple operator with a simple CRD, completing resources by default.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.13 as builder
+FROM golang:1.15 as builder
```
ikarpovich, line 2: Why not 1.16?

api/v1alpha1/progressiverollout_types.go
```diff
@@ -28,18 +30,68 @@ type ProgressiveRolloutSpec struct {
...
 	// Important: Run "make" to regenerate code after modifying this file
 
-	// Foo is an example field of ProgressiveRollout. Edit ProgressiveRollout_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// SourceRef references the resource, example an ApplicationSet, which owns ArgoCD Applications
+	//+kubebuilder:validation:Required
+	SourceRef corev1.TypedLocalObjectReference `json:"sourceRef"`
+	// Stages reference a list of Progressive Rollout stages
+	//+kubebuilder:validation:Optional
+	Stages []*ProgressiveRolloutStage `json:"stages,omitempty"`
```
ikarpovich, line 38: Why is this a reference?

controllers/progressiverollout_controller_test.go
```diff
@@ -0,0 +1,93 @@
...
+			ObjectMeta: metav1.ObjectMeta{Name: "argocd"},
+		}
+		err := k8sClient.Create(context.Background(), &namespace)
+		Expect(err).NotTo(HaveOccurred(), "failed to create test namespace")
+	})
+
+	// We should replace this test with better ones as we deploy the controller
+	It("should reconcile", func() {
+		pr := deploymentskyscannernetv1alpha1.ProgressiveRollout{
+			ObjectMeta: metav1.ObjectMeta{Name: "go-infrabin", Namespace: "argocd"},
```
ikarpovich, line 41: Does the reference to go-infrabin have to be here?

api/v1alpha1/progressiverollout_types.go
sledigabel: I'm guessing it's because the subsequent methods for ProgressiveRolloutStage are ref-based as they will change over time.
@maruina is this your understanding as well?

controllers/progressiverollout_controller_test.go
maruina: No, I just had to invent a name, so maybe having a standard test name?

ikarpovich: IMO it would be better to define a standard test name and stick to it, reference to something existing may create confusion.

api/v1alpha1/progressiverollout_types.go
maruina: No I think it's a good point, I don't think it has to be a reference. This object will store the information and it's read-only. I'll fix that.

sledigabel (dismissed):

ikarpovich (approved):
LGTM, thanks

Dockerfile
maruina: Because it was released 2 days ago and I didn't know :)
Do you think we should update now or we can postpone it?

maruina: go 1.16 is not in `brew` yet, so I will defer the update

sledigabel (approved):

PR #18, opened 2021-02-19
Add Application and Secret watches

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/5

- Add `defer GinkgoRecover()` when starting `k8sManager` so we can see the real issue. Instead of a Ginkgo panic, we can see the real issue. First screenshot is without the `defer`.
image
image

- Add ArgoCD Application CRD to fix the issue above
- Add scripts to import ArgoCD Application package 
- Use recommended Ginkgo [CI settings](http://onsi.github.io/ginkgo/#ginkgo-and-continuous-integration)
- Restructure test to allow multiple `Describe` block
- Randomize namespace creation for each `Describe` block
- Add test for checking if a secret is an ArgoCD cluster
- Add tests for watches
- Build only when we merge into main
- Add RBAC for watching secrets and applications
- Added a predicate to the secret watch to get an event on labels update as per https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.2/pkg/predicate#LabelChangedPredicate
- Create utils package for internal common utilities


controllers/progressiverollout_controller_test.go
```diff
@@ -58,11 +64,50 @@ var _ = Describe("ProgressiveRollout Controller", func() {
...
+						Name:       "single-stage-appset",
+						UID:        uuid.NewUUID(),
+					}},
+				},
+				Spec: argov1alpha1.ApplicationSpec{},
+			}
+			Expect(k8sClient.Create(ctx, ownedApp)).To(Succeed())
+
+			requests := reconciler.requestsForApplicationChange(ownedApp)
+			Expect(len(requests)).To(Equal(1))
```
maruina, line 94: This is a flaky test because sometimes `len(requests)` is 2. I need to investigate why and I suspect because I'm doing something wrong with Ginkgo.

maruina: Solved the flakyness by assigning the ownership only in the required test.

maruina:
One action is failing because we can't build the docker image because the `internal` package is part of this PR and is not yet present in the github repo: https://github.com/Skyscanner/argocd-progressive-rollout/runs/1945265782?check_suite_focus=true#step:9:110

Any ideas about how to fix it?

api/v1alpha1/progressiverollout_types.go
```diff
@@ -90,6 +90,16 @@ func (in *ProgressiveRollout) NewStatusCondition(t string, s metav1.ConditionSta
 	}
 }
 
+// HasOwnerReference returns true if the ProgressiveRollout object has a reference to one of the owners
+func (in *ProgressiveRollout) HasOwnerReference(owners []metav1.OwnerReference) bool {
```
sledigabel, line 94: I think I got slightly confused with the name of the function. It implies that the ProgressiveRollout has an attribute called OwnerRefrence and it would have one or not.
I'd rename it `IsReferredByOwners` or something like that?

Makefile
```diff
@@ -15,7 +15,12 @@ all: manager
 
 # Run tests
 test: generate fmt vet manifests
-	go test -v -covermode=atomic -coverprofile=coverage.out -race ./...
+	ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
```
Smirl, line 18: Does this give better output than the trash built in?

internal/utils.go
```diff
@@ -0,0 +1,17 @@
+package internal
```
sledigabel, line 1: that file is a bit weird.
it's called utils but it's in `internal` and it has helper functions.
It'd rather we move the constants that will be reused in a separate package, and rename the package `utils`?

controllers/progressiverollout_controller.go
```diff
@@ -61,5 +69,93 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
 func (r *ProgressiveRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&deploymentskyscannernetv1alpha1.ProgressiveRollout{}).
+		Watches(
+			&source.Kind{Type: &argov1alpha1.Application{}},
+			handler.EnqueueRequestsFromMapFunc(r.requestsForApplicationChange)).
```
sledigabel, line 74: curious to see what that line does

api/v1alpha1/progressiverollout_types.go
Smirl, line 94: Should this be some thing like `IsOwnedBy`? I initially read it as if the `ProgressiveRollout` had any owner reference on it.

controllers/progressiverollout_controller.go
```diff
@@ -61,5 +69,93 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+func (r *ProgressiveRolloutReconciler) requestsForApplicationChange(o client.Object) []reconcile.Request {
+
+	/*
+		We trigger a reconciliation loop on an Application event if:
+		- the Application owner is referenced by a ProgressiveRollout object
+	*/
+
+	var requests []reconcile.Request
+	var list deploymentskyscannernetv1alpha1.ProgressiveRolloutList
+	ctx := context.Background()
```
sledigabel, line 89: I was wondering about this one. Will we _always_ use Background() for `ctx` or would there be cases where it would be interesting to cascade timeouts?

config/crd/bases/argoproj.io_application.yaml
```diff
@@ -0,0 +1,1709 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
Smirl, line 1: Strange that this is here.
Does this mean that when you deploy this it also deploys the Application CRD?

controllers/progressiverollout_controller.go
```diff
@@ -61,5 +69,93 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			handler.EnqueueRequestsFromMapFunc(r.requestsForApplicationChange)).
+		Watches(
+			&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(r.requestsForSecretChange)).
 		Complete(r)
 }
+
+func (r *ProgressiveRolloutReconciler) requestsForApplicationChange(o client.Object) []reconcile.Request {
+
+	/*
+		We trigger a reconciliation loop on an Application event if:
```
Smirl, line 83: Makes it seem like we are reconciling the Application not the ProgressiveRollout (although that should be obvious as this is the `ProgressiveRolloutReconciler`

Smirl, line 83: Makes it seem like we are reconciling the Application not the ProgressiveRollout (although that should be obvious as this is the `ProgressiveRolloutReconciler`

controllers/progressiverollout_controller.go
```diff
@@ -61,5 +69,93 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+
+	if !internal.IsArgoCDCluster(s.GetLabels()) {
+		return nil
+	}
+
+	if err := r.List(ctx, &prList); err != nil {
+		r.Log.Error(err, "failed to list ProgressiveRollout")
+		return nil
+	}
+	if err := r.List(ctx, &appList); err != nil {
```
Smirl, line 144: ❓ This function must be magic? It looks at the type of the objected passed it and lists thoses? 🤯

controllers/progressiverollout_controller.go
```diff
@@ -61,5 +69,93 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+		r.Log.Error(err, "failed to list Application")
+		return nil
+	}
+
+	for _, pr := range prList.Items {
+		for _, app := range appList.Items {
+			if app.Spec.Destination.Server == string(s.Data["server"]) && pr.HasOwnerReference(app.GetOwnerReferences()) {
```
Smirl, line 151: As this is a loop. If I have 2 apps watched by 1 pr, or 2 apps watched by 2 pr, will the output on this function be longer than 1?

i.e. should we deduplicate this list?

controllers/progressiverollout_controller.go
```diff
@@ -61,5 +69,93 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			}})
+		}
+	}
+
+	return requests
+}
+
+func (r *ProgressiveRolloutReconciler) requestsForSecretChange(o client.Object) []reconcile.Request {
+
+	/*
```
sledigabel, line 117: Just so I'm clear; is the logic of reconciling after a secret existing because we want to notify the current deployments that a cluster has been marked as drained?
Just trying to understand the impact, would that effectively trigger a deployment? Idempotency should mean no change, but just concious that it could mean _every single_ progressiverollout will trigger at the same time?

There might be something I'm not getting here.

internal/utils.go
```diff
@@ -0,0 +1,17 @@
...
+const (
+	ArgoCDSecretTypeLabel   = "argocd.argoproj.io/secret-type"
+	ArgoCDSecretTypeCluster = "cluster"
+	AppSetKind              = "ApplicationSet"
+	AppSetAPIGroup          = "argoproj.io/v1alpha1"
+)
+
+func IsArgoCDCluster(labels map[string]string) bool {
+	val, ok := labels[ArgoCDSecretTypeLabel]
+	if val == ArgoCDSecretTypeCluster && ok {
```
Smirl, line 12: perhaps just return the condition without the if/else?

Smirl (changes requested):

internal/utils.go
```diff
@@ -0,0 +1,17 @@
+package internal
+
+const (
+	ArgoCDSecretTypeLabel   = "argocd.argoproj.io/secret-type"
+	ArgoCDSecretTypeCluster = "cluster"
+	AppSetKind              = "ApplicationSet"
+	AppSetAPIGroup          = "argoproj.io/v1alpha1"
+)
+
+func IsArgoCDCluster(labels map[string]string) bool {
```
sledigabel, line 10: I'm not super clear on what this function achieves

Makefile
maruina: Yes, it also gives you an idea about how many tests you are doing with a nice progress bar

api/v1alpha1/progressiverollout_types.go
maruina: Fixed

controllers/progressiverollout_controller.go
maruina: Do you have a suggestion about how to better phrase it?

controllers/progressiverollout_controller.go
maruina: The object type is `somethingList`. It works the same as https://github.com/Skyscanner/argocd-progressive-rollout/blob/main/controllers/progressiverollout_controller.go#L46

controllers/progressiverollout_controller.go
maruina: You can't have 2 Apps watched by 2 PRs, because this mean we have 2 PRs watching the same AppSet and then which PR "wins" in deciding how to update the Applications? We probably need a validation hook for this

Regarding the 2 Apps watched by 1 PR that happens all the time, because an AppSet can generate multiple applications.

Is this more clear? I think I don't get the issue you are pointing at.

internal/utils.go
maruina: Ah, good point. Fixed

.github/workflows/test-build.yaml
```diff
@@ -15,11 +17,6 @@ jobs:
         with:
           go-version: ^1.15
 
-      - name: Initialize CodeQL
```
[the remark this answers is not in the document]
maruina: I moved in a different file so we can run the job in parallel. See .github/workflows/code-quality.yaml

api/v1alpha1/progressiverollout_types.go
maruina: Yeah, done that as per @Smirl comment

internal/utils.go
maruina: My understanding looking at https://github.com/golang-standards/project-layout is that you put in internal things you don't want to expose as public package, so all the utility I think are related to this project but I don't want other people to import.

They do the same here for example: https://github.com/fluxcd/kustomize-controller/tree/main/internal/sops

Do you think we should expose all those utilities?

controllers/progressiverollout_controller.go
maruina: Did you have a look at https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/handler#EnqueueRequestsFromMapFunc? Anything I can add to make it more clear?

controllers/progressiverollout_controller.go
maruina: It's a good question, I don't know know yet.

I don't think I _ever_ seen a controller passing around `ctx`, but that is not a good reason why we shouldn't do it if we think it's a good idea.

controllers/progressiverollout_controller.go
maruina: We trigger a reconciliation loop every time there is a change to a secret object in Lubernetes. This is why we filter using this function, because we don't care about *every* secrets.

Let's say we add a new label to a secret, this will trigger a reconciliation loop for every ProgressiveRollout targeting that cluster.

We could minimize the impact and do something like "don't forward the event if we are *adding* the following `key:value`". In that case we could avoid triggering the reconciliation loop when we add `drained: true` but still trigger it when we remove it or change to `drained: false`.

If that is what you mean, I think I will raise a separate issue to fix that. We might want by default filter out all the labels used in the (future) requeue work.

internal/utils.go
maruina: Looks at the Secret labels and returns true if it has the ArgoCD key and value. We don't care about all the secrets, we just care about the ones storing ArgoCD cluster information.

internal/utils.go
sledigabel: yeah it's more semantics than the layout, I'd name the package utils, in a `utils` directory in the `internal` directory, it would make sense. The import would be `utils`.
I certainly do not think we should expose them.

I'd move the appset constants that will be reused throughout the code base in a separate package. The AppSetKind and APIGroups are not really helpers. This is the last place I would think of to look for this. We could make a new package for it. What do you think? the other two are used locally, happy for them to stay there

internal/utils.go
sledigabel: ok, it makes sense. Worth renaming the function `IsArgoCDClusterSecret`?

Smirl (approved):

internal/utils.go
maruina: I'm not very confused by the name, so if you feel we should rename it I can do that.

internal/utils.go
maruina: Cool, I'll do a separate utils package.
Regarding the constant, do you have any suggestion for where to put them?

config/crd/bases/argoproj.io_application.yaml
maruina: Moved into `hack`

controllers/progressiverollout_controller.go
sledigabel: that is what I meant indeed. Happy to create a separate issue.

internal/utils.go
sledigabel: The function on its own did, but used in context it's actually fine.
```
	if !utils.IsArgoCDCluster(s.GetLabels()) {
		return nil
	}
```

internal/utils.go
maruina: Since `const` is a reserved name, I kept them in utils for the time being, we can move them later.

controllers/progressiverollout_controller.go
maruina: > The question still stands when a new cell is added, in this case the argocd secret would be updated too, does this mean all the progressive rollouts would trigger a deployment?

Only if there are Applications targeting that new cluster, so none until we have Applications able to deploy there.

We can use with https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.8.2/pkg/predicate to create our own filter logic so we send only the events we care about.

controllers/progressiverollout_controller.go
sledigabel: Let's leave it for now, and keep it in mind in case timeouts become a problem.

internal/utils.go
sledigabel: fine, i'll create a quick issue so we don't forget.

controllers/progressiverollout_controller.go
```diff
@@ -73,7 +78,8 @@ func (r *ProgressiveRolloutReconciler) SetupWithManager(mgr ctrl.Manager) error
 			&source.Kind{Type: &argov1alpha1.Application{}},
 			handler.EnqueueRequestsFromMapFunc(r.requestsForApplicationChange)).
 		Watches(
-			&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(r.requestsForSecretChange)).
+			&source.Kind{Type: &corev1.Secret{}}, handler.EnqueueRequestsFromMapFunc(r.requestsForSecretChange),
+			builder.WithPredicates(predicate.Or(predicate.GenerationChangedPredicate{}, predicate.LabelChangedPredicate{}))).
```
sledigabel, line 82: 👍

sledigabel (approved):
LGTM

PR #21, opened 2021-02-22
Add targets selection logic

### LINKED ISSUE #17: Add targets selection logic
Add functions to retrieve the targets selection. We should be able to test it in isolation, outside the reconciliation loop.

maruina:
Fix #17.

Notes
- ~Depends on https://github.com/Skyscanner/argocd-progressive-rollout/pull/18, I will rebase once 18 is merged before marking this PR ready to be reviewed.~
- I didn't add any test yet because we would end up testing the label selector. Tests will come when we start doing the scheduling bit.
- Fix Docker build by adding missing `internal/`


controllers/progressiverollout_controller.go
```diff
@@ -180,3 +197,26 @@ func (r *ProgressiveRolloutReconciler) requestsForSecretChange(o client.Object)
 
 	return requests
 }
+
+// GetTargetClusters returns a list of ArgoCD clusters matching the provided label selector
+func (r *ProgressiveRolloutReconciler) GetTargetClusters(selector metav1.LabelSelector) (corev1.SecretList, error) {
```
sledigabel, line 202: what is the logic behind returning a list of secrets here instead of a list of clusters?
argo-cd has a specific object `Cluster`, wouldn't that make more sense to use that? It would make the validation with the argocd project easier as well down the line, and manipulating the same types across the ecosystem is a bit more future proof too.

https://github.com/argoproj/argo-cd/blob/12ec99ad01a945e045225e5c239f8ee203e8139a/util/db/cluster.go#L80-L100
https://github.com/argoproj/argo-cd/blob/12ec99ad01a945e045225e5c239f8ee203e8139a/pkg/apis/application/v1alpha1/types.go#L1104-L1126

sledigabel: they don't seem to have a selector though :-(

maruina: Because from the Kubernetes point of view they *are* secrets, the `cluster` object doesn't exist.

maruina: It is also the same strategy used by the cluster generator in the ApplicationSet, as per https://github.com/argoproj-labs/applicationset/blob/master/pkg/generators/cluster.go

sledigabel: OK fine, thanks for pointing me the example on appsets.
we can still unmarshall the secrets and create/reuse a specific cluster type later.

sledigabel (approved):
LGTM, adds a cluster list getter based on a selector.

PR #34, opened 2021-03-02
Add Helm chart

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/13

- Add an Helm3 chart but still compatible with Helm2.
- Use [helm-docs](https://github.com/norwoodj/helm-docs) to automatically generate the chart `README.md` from a template
- Add helm-docs as git pre-commit hook
- Use official Helm actions to lint and test charts
- Use official Helm actions to store the chart in this repository


maruina:
I'm not sure what I can do to make the checks passing

maruina:
Done a new push to see if we can have all green.

charts/argocd-progressive-rollout/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "argocd-progressive-rollout.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create }}
+{{- default (include "argocd-progressive-rollout.fullname" .) .Values.serviceAccount.name }}
+{{- else }}
+{{- default "default" .Values.serviceAccount.name }}
```
nebojsa-prodana, line 60: Is this safe to do, ie have the chart add permissions to the default user if no user is specified?

charts/argocd-progressive-rollout/templates/deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+            - /manager
+            - --metrics-addr={{ .Values.args.metricsAddr }}
+            {{- if or (gt ( .Values.replicaCount | int64) 1) .Values.args.enableLeaderElection }}
+            - --enable-leader-election=true
+            {{- end }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: http-metrics
+              containerPort: {{ (split ":" .Values.args.metricsAddr)._1 }}
```
nebojsa-prodana, line 42: Would it be better to have two params for hostname and port in the values file instead of having to split here?

charts/argocd-progressive-rollout/templates/psp.yaml
```diff
@@ -0,0 +1,57 @@
...
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+    - '*'
+---
+kind: ClusterRole
```
nebojsa-prodana, line 30: Should we be granting a cluster wide role or scope it to a namespace instead?

charts/argocd-progressive-rollout/crds/crd-progressiverollout.yaml
```diff
@@ -0,0 +1,230 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
```
nebojsa-prodana, line 2: This is just a copy from /config right? I do not know what the best practice is, but can we reference the file from there directly instead so that we do not have to worry about them going out of sync? Or could the helm charts be stored inside of config/crd without messing up with kustomize?

charts/argocd-progressive-rollout/templates/_helpers.tpl
maruina: Yes, because if you don't specify a service account name, Kubernetes will assign you one called `default` in the same namespace. See https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server

charts/argocd-progressive-rollout/templates/deployment.yaml
maruina: I was doing this because then we have 1:1 match between the chart arguments and the controller flags. I'd prefer keep it this way unless you have a strong objection.

charts/argocd-progressive-rollout/templates/psp.yaml
maruina: I was porting the Kustomize version to Helm as per https://github.com/Skyscanner/argocd-progressive-rollout/blob/main/config/rbac/role.yaml
When you use the kubebuilder annotation it creates a ClusteRole as per https://book.kubebuilder.io/reference/markers/rbac.html

Do you have any concern?

charts/argocd-progressive-rollout/crds/crd-progressiverollout.yaml
maruina: It's a copy&paste. We can't reference the file because in Helm they need to be YAML file as per https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

It's best practice to keep them separate, so we can't add the helm chart inside `config/crd`. I think it would confuse people having an Helm chart stored under a Kustomize subdirectory. What do you think?

charts/argocd-progressive-rollout/templates/deployment.yaml
nebojsa-prodana: No, not a strong objection - I saw that in other places, ie for image we have separate fields so I was wondering why this would be different.

charts/argocd-progressive-rollout/templates/psp.yaml
nebojsa-prodana: Just trying to see if we can give the least permissions as possible to the operator. We will be monitoring a single argocd namespace, right so technically we don't need a ClusterRole? I am fine with keeping it as a ClusterRole if that is what kubebuilder recommends.

charts/argocd-progressive-rollout/crds/crd-progressiverollout.yaml
nebojsa-prodana: I know that we said that we do not expect the crd definition to change often, so I guess it is fine to keep it as a copy for now. I agree that it would be awkward to have a chart inside of a kustomize subdir. Maybe in the future we can extend `make generate` or some other make target just to make sure that these are kept in sync?

nebojsa-prodana (approved):
LGTM!

charts/argocd-progressive-rollout/crds/crd-progressiverollout.yaml
sledigabel: this is a good idea.

sledigabel: or a symlink?

charts/argocd-progressive-rollout/templates/psp.yaml
maruina: Since the RBAC is generated by the kubebuilder itself, I don't see a way to change it looking at https://book.kubebuilder.io/reference/markers/rbac.html

charts/argocd-progressive-rollout/crds/crd-progressiverollout.yaml
maruina: Yeah, that's a good idea.

PR #37, opened 2021-03-04
Publish helm chart on main

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/pull/34 by

- using the correct chart directory
- publishing the chart when we merge into main


sledigabel (approved):
LGTM, pushes the new chart at each merge to main.

PR #38, opened 2021-03-04
Fix helm release

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/pull/37 by

- removing `charts_dir` since it already defaults to `charts` as per https://github.com/helm/chart-releaser-action/blob/master/cr.sh#L30
- fix `version` as it was missing a `v`
image


sledigabel (approved):
LGTM, fixes the action so that it uses the right version of chart releaser

PR #39, opened 2021-03-04
Update the Helm Readme

maruina:
Tested locally now we have a successful chart release.

```
❯ helm repo add argocd-progressive-rollout https://skyscanner.github.io/argocd-progressive-rollout/
"argocd-progressive-rollout" has been added to your repositories

❯ helm upgrade -i apr argocd-progressive-rollout/argocd-progressive-rollout --namespace argocd --devel
Release "apr" does not exist. Installing it now.
W0304 11:50:23.129748   85648 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
NAME: apr
LAST DEPLOYED: Thu Mar  4 11:50:25 2021
NAMESPACE: argocd
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

Note that I had to use `--devel` because our current release is `0.1.0-prealpha`. That will not be necessary once we do a first release, so I don't think they should go in "official" instructions.


sledigabel (approved):
LGTM, updates the doc with upgrade instead of install (best practice) and bump version

PR #42, opened 2021-03-10
Create scripts for bootstrapping the env

nebojsa-prodana:
Implements https://github.com/Skyscanner/argocd-progressive-rollout/issues/28

It is not invoked via make file for now, I feel that the makefile is full of targets atm making it a bit difficult to follow, but I can add it there and update the readme accordingly if that'd be preffered.


hack/setup-dev.sh
```diff
@@ -0,0 +1,55 @@
...
+source $root/dev-functions.sh
+
+# Create the control cluster
+kind delete cluster --name argocd-control-plane
+kind create cluster --name argocd-control-plane
+kubectl create namespace argocd
+
+# Setup argocd
+kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml
+kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj-labs/applicationset/v0.1.0/manifests/install.yaml
```
sledigabel, line 16: let's make the appset version an environment variable?

hack/add-cluster.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -e
+
+root=$(dirname "${BASH_SOURCE[0]}")
+bash $root/install-dev-deps.sh
```
sledigabel, line 5: I would just leave that script independent.
People can use it if they want to, but don't have to install things through the script. I understand the script checks whether things are already installed, but i'd prefer the tool to tell me something is missing and suggest a script to install it rather than doing it for me

I'm a huge brew user for instance, and would prefer to track it there rather than curling it from github

hack/add-cluster.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
sledigabel, line 1: doesn't seem like this script adds a lot more functionalities than calling out register_argocd_cluster?
Is this needed as a fully separated script?

hack/setup-dev.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+set -e
+
+root=$(dirname "${BASH_SOURCE[0]}")
+
+bash -e $root/install-dev-deps.sh
```
sledigabel, line 6: as discussed before, i'd remove this and replace perhaps with a check to make sure all the required tools are present

hack/setup-dev.sh
```diff
@@ -0,0 +1,55 @@
...
+kind delete cluster --name argocd-control-plane
+kind create cluster --name argocd-control-plane
+kubectl create namespace argocd
+
+# Setup argocd
+kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj/argo-cd/stable/manifests/install.yaml
+kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj-labs/applicationset/v0.1.0/manifests/install.yaml
+
+# Login as admin and change def pass
+sleep 10
```
sledigabel, line 19: what is the sleep step for?

hack/setup-dev.sh
```diff
@@ -0,0 +1,55 @@
...
+kubectl apply -f $root/dev/control-plane.yml
+
+# Create additional clusters to server as deployment targets
+register_argocd_cluster "prc-cluster-1" true
+register_argocd_cluster "prc-cluster-2" true
+
+# Generate token for the new user
+token=$(retry_argocd_exec "$argocdlogin >/dev/null && argocd account generate-token --account prc")
+serverip=$(kubectl get service -n argocd argocd-server -o=jsonpath='{.spec.clusterIP}')
+sleep 5
```
sledigabel, line 44: same comment as above

hack/add-cluster.sh
nebojsa-prodana: Yeah, I was thinking among the same lines. I've decided to keep it like this for simplicity and have opted for curling stuff instead of relying on brew to make it more cross platformy. In the end one of my intention is to leverage these scripts for github actions e2e testing against actual kind clusters.

The other reason why I used curls instead of brew is because I got burned on installing the brew version of kubebuilder which doesn't come with etcd so the BDD tests were failing.

I propose that:
* We separate the scripts and add instructions on how to run them
* We keep curling for now for simplicity sake and rethink whether we want to offer a brew alternative for OSX once we have resolved the hurdles of installing this in the CI env.

hack/add-cluster.sh
nebojsa-prodana: It is just a convenience script. You'd have to source the functions script and call it instead. It is easier and less polluting of your local shell to just invoke a separate script if you want to add another cluster to your setup.

hack/setup-dev.sh
nebojsa-prodana: Sometimes the object is not immediately retrievable in kubernetes so this makes sure that the watch on the line below doesn't fail because the object has not yet been created.

hack/setup-dev.sh
nebojsa-prodana: There are some race conditions of the exec process writing to stdout that might corrupt our echoing of token and server ip below. We can probably find the right level of redirecting stuff to /dev/null to avoid this, but I'd rather provide the ability to bootstrap your own env as soon as possible and iterate on resolving this sort of issues later.

hack/setup-dev.sh
nebojsa-prodana: Yeah, we should probably do the same for argocd as well. As what is stable can easily change underneath us.

maruina:
Had a chat with @nebojsa-prodana about having a style guide since bash can very (too?) opinionated. We thought about using:
- https://google.github.io/styleguide/shellguide.html for code style
- https://pkg.go.dev/mvdan.cc/sh/v3 for formatting
- https://www.shellcheck.net/ for static analysis

We should probably add a couple of github actions for shell as well

hack/add-cluster.sh
sledigabel: That sounds good; i wasn't suggesting by all means enforce brew or anything. My point was to not force the installation.

hack/setup-dev.sh
sledigabel: i've been there :-) could this be simplified / clarified by a `kubectl wait`?

hack/add-cluster.sh
sledigabel: Would it make more sense to invoke this script in the `setup.sh` then rather than using the `register_argocd_cluster`?

hack/setup-dev.sh
```diff
@@ -0,0 +1,67 @@
...
+# Generate token for the new user
+token=$(retry_argocd_exec "$argocdlogin >/dev/null && argocd account generate-token --account prc")
+serverip=$(kubectl get service -n argocd argocd-server -o=jsonpath='{.spec.clusterIP}')
+
+# retry_argocd_exec can race with outputting to stdout with the rest of this script
+# so sleep for a bit to make sure that the token and server ip output doesn't get mangled
+sleep 5
+
+echo "All done"
+echo "Token:$token"
```
sledigabel, line 60: I suggest you store the token and the serverip in a separate file or something easy to source for subsequent uses

go-mod-hack.sh
```diff
@@ -1,21 +1,24 @@
-#!/bin/sh
+#!/bin/bash
```
maruina, line 1: This script was a copy&paste from a github issue. Did you check if it's still working as expected?

go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/Skyscanner/argocd-progressive-rollout
 go 1.15
 
 require (
-	github.com/argoproj/argo-cd v0.8.1-0.20210218202601-6de3cf44a4cb
+	github.com/argoproj/argo-cd v0.8.1-0.20210218100039-a4ee25b59d8d
```
maruina, line 6: If you are only adding some scripts, why is go.mod changing?

hack/add-cluster.sh
```diff
@@ -0,0 +1,17 @@
...
+set -e
```
maruina, line 2: I would add `-u` since we don't do any pipe (so we can omit `-o pipefail`).

hack/dev/perms.yml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
```
maruina, line 1: Maybe `rbac.yaml`?

hack/dev/test-appset.yml
```diff
@@ -0,0 +1,35 @@
...
+          url: https://prc-cluster-1-control-plane:6443
+        - cluster: kind-prc-cluster-2
+          url: https://prc-cluster-2-control-plane:6443
+  template:
+    metadata:
+      name: 'goinfra-{{cluster}}'
+    spec:
+      project: default
+      source:
+        repoURL: https://github.com/maruina/appset-poc.git
```
maruina, line 22: Should we point this to this repository instead of relying on a third party one?

hack/install-argocd-application.sh
```diff
@@ -1,5 +1,8 @@
-#/bin/sh
+#!/bin/bash
 set -e
 
-../go-mod-hack.sh v1.20.1
+# root=$(dirname "${BASH_SOURCE[0]}")
+
+bash go-mod-hack.sh v1.20.1
 go get github.com/argoproj/argo-cd@a4ee25b
+touch hack/.hack.argocd.installed
```
maruina, line 8: Do we need to `.gitignore` this file?

hack/install-dev-deps.sh
```diff
@@ -0,0 +1,55 @@
...
+if ! [ -x "$(command -v pre-commit)" ]; then
+	[ "$check_only" == true ] && err "pre-commit is not installed. Run bash hack/install-dev-deps.sh to install." && exit 1
+	pip install --index https://pypi.python.org/simple pre-commit
+fi
+
+if [ ! -d "/usr/local/kubebuilder" ]; then
+	[ "$check_only" == true ] && err "kubebuilder is not installed. Run bash hack/install-dev-deps.sh to install." && exit 1
+
+	# download kubebuilder and extract it to tmp
+	curl -L "https://go.kubebuilder.io/dl/2.3.1/${os}/${arch}" | tar -xz -C /tmp/
```
maruina, line 24: Should we add the kubebuilder version as variable?

hack/install-dev-deps.sh
```diff
@@ -0,0 +1,55 @@
...
+
+	# Because of https://github.com/argoproj/argo-cd/issues/4055) we can't just run `go get github.com/argoproj/argo-cd`.
+	bash -xe "$root"/install-argocd-application.sh
+fi
+
+# Install kind
+if [ ! -f "$gopath/bin/kind" ]; then
+	[ "$check_only" == true ] && err "kind is not installed. Run bash hack/install-dev-deps.sh to install." && exit 1
+
+	curl -Lo ./kind "https://kind.sigs.k8s.io/dl/v0.10.0/kind-$os-$arch"
```
maruina, line 43: Same for kubebuilder, should we have this as parameter at the top?

hack/setup-dev.sh
DiogoMCampos: What I did right after running the script was storing it in a file, indeed. Perhaps we could do that add have said file in `.gitignore` as that looks like the user-friendly way to source these.

maruina (approved):
Looks good so far.
Looking at the style guide I would probably use `[[` instead of `[` because of https://google.github.io/styleguide/shellguide.html#test----and---

hack/install-dev-deps.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+set -e
```
maruina, line 2: Here I would add `-uo pipefail`

DiogoMCampos:
For the record, ran this in `5.10.19-1-MANJARO` and these were my notes:
 - `go` and `kubectl` are pre-requisites to these scripts;
 - `~/go/bin` didn't exist on my system (freshly installed Go through AUR);
 - `curl` didn't have permissions to download the `argocd` CLI into `/usr/local/bin` in my system - this could be related to my specific installation, and is not a problem on my OSX;
 - I had to run `export PATH=$PATH:$(go env GOPATH)/bin` (temporary solution, but I'll add it permanently) to get `zsh` to find the `kind` command.
 
 Those are minor things that are mostly user-action and we don't need to solve them all. The main logic worked as expected.

hack/setup-dev.sh
sledigabel: yeah that sounds right. as discussed with @nebojsa-prodana i think we need to make the steps to having a cluster and accessing the command line and UI as easy as possible.

README.md
```diff
@@ -64,11 +64,34 @@ See [CONTRIBUTING.md](./CONTRIBUTING.md)
...
 
-### Update ArgoCD Application API package
+```
+bash hack/install-dev-deps.sh
+bash hack/setup-dev.sh [argocd-version] [appset-version]
+```
+
+this will install all the dependencies (`pre-commit`, `kubebuilder`, `argocd`, `kind`) and it will install the correct version of ArgoCD Application API package for you. If you omit `argocd-version` and/or `appset-version` it will default to the latest stable/tested versions of ArgoCD and Appset controller.
+
+After running the script, you will have 3 kind clusters created locally:
```
maruina, line 76: Should we differentiate between instructions to run the controller locally with the kubebuilder api server and etcd, and a more e2e environment with a real kubernetes cluster, argocd and applicationset?

For developing the controller should be enough what we get with kubebuilder.

hack/setup-dev.sh
sledigabel: @nebojsa-prodana i just had a look at kind tonight. i come from a minikube background (backed by a VM with its own IP), and just realised that kind is backed by the funky network layer from Docker. Bottomline is `NodePort` is a terrible idea. Please forget what I said about that.

I still think having a file is needed with the serverip and the token, and perhaps a kubectl port-forward shortcut would be handy.

go-mod-hack.sh
nebojsa-prodana: yep, i made sure it is working still :)

hack/install-argocd-application.sh
nebojsa-prodana: yes, it is just a flag file to make sure we don't have to go through installing the go-mod-hack over and over again.

go.mod
nebojsa-prodana: I think it moved the needle when i ran go-mod-hack, I can omit it from the pr if you want, but it will probably be changing all the time.

hack/dev/test-appset.yml
nebojsa-prodana: Can this be left for latter iterations? This PR is already getting a bit large, and I wouldn't want to add additional files to it at the moment.

hack/dev-functions.sh
```diff
@@ -0,0 +1,95 @@
...
+
+	clustername=$1
+	recreate=${2:-false}
+
+	if [[ "$recreate" == true ]]; then
+		kind delete cluster --name "$clustername"
+	fi
+
+	# TODO: move these into a function
+	argoserver=$(kubectl get pods -n argocd -l app.kubernetes.io/name=argocd-server -o name | cut -d'/' -f 2)
```
sledigabel, line 33: `kubectl get pods -n argocd -l app.kubernetes.io/name=argocd-server -o jsonpath='{.items..metadata.name}'`

hack/dev-functions.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+set -eu
+
+function err() {
+	echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $*" >&2
+}
+
+function retry_argocd_exec() {
+	argoserver=$(kubectl get pods -n argocd -l app.kubernetes.io/name=argocd-server -o name | cut -d'/' -f 2)
```
sledigabel, line 9:
```suggestion
	argoserver=$(kubectl get pods -n argocd -l app.kubernetes.io/name=argocd-server -o jsonpath='{.items..metadata.name}')
```

hack/dev-functions.sh
```diff
@@ -0,0 +1,95 @@
...
+	kubectl config use-context "$prevcontext" >/dev/null
+
+	docker stop "$socat_container" >/dev/null || err "No proxy container running"
+	docker rm "$socat_container" >/dev/null || err "No proxy container running"
+
+	# Kind isn't kind when it comes to exposing additional nodeports
+	# since they also need to be mapped to docker
+	# However, we can run socat to proxy a tcp socket without having to meddle with
+	# the kind controlplane container. See: https://github.com/kubernetes-sigs/kind/issues/99#issuecomment-456184883
+	docker run \
```
sledigabel, line 70: OK, that's pretty smart 👍

README.md
nebojsa-prodana: I think I've addressed this now. Let me know what you think.

hack/dev-functions.sh
```diff
@@ -0,0 +1,95 @@
...
+		echo "ARGOCD_PASSWORD=prc"
+		echo "ARGOCD_INCLUSTER_ADDRESS=$serverip"
+		echo "ARGOCD_LOCAL_ADDRESS=https://localhost:$argoserver_nodeport"
+	} >.env.local
+
+	# Create a secret storing the token and in-cluster ip
+	kubectl delete secret generic -n argocd prc-controller-secret || err "Secret not found. Creating.."
+	kubectl create secret generic -n argocd prc-controller-secret --from-literal="token=$token" --from-literal="serverip=$serverip" >/dev/null
+
+	echo "https://localhost:$argoserver_nodeport"
```
sledigabel, line 94:
```suggestion
	open "https://localhost:$argoserver_nodeport"
```
Could be worth opening it straight away?

hack/dev-functions.sh
```diff
@@ -0,0 +1,95 @@
...
+
+	argocd login --insecure --username prc --password prc localhost:"$argoserver_nodeport" >/dev/null
+
+	{
+		echo "ARGOCD_TOKEN=$token"
+		echo "ARGOCD_USERNAME=prc"
+		echo "ARGOCD_PASSWORD=prc"
+		echo "ARGOCD_INCLUSTER_ADDRESS=$serverip"
+		echo "ARGOCD_LOCAL_ADDRESS=https://localhost:$argoserver_nodeport"
+	} >.env.local
```
sledigabel, line 88: Should we cat it?

hack/dev-functions.sh
nebojsa-prodana: I need this to be the return value of function, so I have to `echo` it in here, but I can add `open` to the places where it is being called.

hack/dev-functions.sh
nebojsa-prodana: Not sure what advantages would that give us? Is it just something that you would find more readable/idiomatic?

sledigabel: the logic is that you can just login straight up without having to dump the file again.

sledigabel: it doesn't have to be, this is pretty minor

sledigabel (approved):
LGTM, sets up a couple of clusters using KIND and registers them in argocd, exposes the port.
Sets up the environment if you're missing some local tools.

PR #43, opened 2021-03-11
Fix removeAnnotationFromApps to use value semantic

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/33


sledigabel (approved):
Nice, LGTM, fixes the function definition and use a slice.
Adds a test for it too.

Issue #33, opened 2021-03-01
Discuss pointer vs value semantics, agree and refactor


sledigabel:
The operator has a mix of pointer semantics vs value semantics.
More than the decision about _why_, we need to be consistent and not flip flop between one or the other.

https://github.com/Skyscanner/argocd-progressive-rollout/pull/25/files#diff-b1a5cf22f2be07477047c7c268c9a2077f75c3fdadc285eb862dab8bc3aaad31R293
https://github.com/Skyscanner/argocd-progressive-rollout/pull/25/files#diff-c3753ff27e79b7da6571efc4fffc45e729b031a39337eec9b22a04425bbd69f0R12

https://tour.golang.org/methods/8
https://www.ardanlabs.com/blog/2017/06/design-philosophy-on-data-and-semantics.html

maruina:
We discussed and agree to keep:
- pointer semantic for the controller and reconciler
- value semantic for everything else

So we need to change only this method: https://github.com/Skyscanner/argocd-progressive-rollout/blob/main/controllers/progressiverollout_controller.go#L293

PR #44, opened 2021-03-16
Send Application Sync request to Argo CD Server

DiogoMCampos:
Issue link: https://github.com/Skyscanner/argocd-progressive-rollout/issues/22

Opted to use the code behind the CLI, authenticating against the server based on secrets read from a local file. For the initial implementation, we opted to use Insecure.

### What this does
- Reads configuration from two files.
- Creates a client to the argocd server.
- Sends a sync request for every app that's scheduled to do so.
- Sets up secrets to run controller locally as part of the set up scripts.

### What this doesn't do
- Doesn't handle certificates to connect securely to the argocd server;
- Doesn't do anything on sync failure;

There's a couple of things missing that I intend to work on still, namely understanding what I need to change to the definitions/Makefile to use and mount the secrets as files.


controllers/progressiverollout_controller.go
```diff
@@ -92,11 +94,21 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 
 		for _, s := range scheduledApps {
-			// TODO: add sync method here
 			r.Log.Info("syncing app", "app", s)
+
+			_, err := r.syncApp(s.Name)
+
+			if err != nil {
+				if strings.Contains(err.Error(), "another operation is already in progress") {
+					log.Info("another operation is already in progress", "app", s)
```
DiogoMCampos, line 106: I assume this means we proceed as normal - it's not an error _per se_ - so I don't return here.

controllers/progressiverollout_controller.go
```diff
@@ -92,11 +94,21 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+				} else {
+					log.Error(err, "unable to sync app", "app", s)
+					return ctrl.Result{}, err
```
DiogoMCampos, line 109: Is this all we want to do? Do we want to somehow signal the stage as failed? Or is that future work?

controllers/progressiverollout_controller_test.go
```diff
@@ -22,6 +25,26 @@ const (
 	interval = time.Millisecond * 10
 )
 
+type mockArgoCDAppClientCounter struct {
+	appsSynced []string
+	m          sync.Mutex
```
DiogoMCampos, line 30: The `mutex` is needed because the mocked `Sync` access the `appsSynced` to append stuff and there is the chance of a data race with the `Eventually()` in the test using this.

hack/dev-functions.sh
```diff
@@ -87,6 +87,13 @@ function local_argocd_login() {
 		echo "ARGOCD_LOCAL_ADDRESS=https://localhost:$argoserver_nodeport"
 	} >.env.local
 
+	if [ ! -d .prcconfig ]; then
```
DiogoMCampos, line 90: This sets up a config folder locally so we can `make run` and use it to connect to the `argocd-server` in `kind`.

controllers/progressiverollout_controller.go
```diff
@@ -92,11 +94,21 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 
 		// Get the Applications to update
-		scheduledApps := scheduler.Scheduler(apps, stage)
+		scheduledApps := scheduler.Scheduler(outOfSyncApps, stage)
 
 		for _, s := range scheduledApps {
-			// TODO: add sync method here
 			r.Log.Info("syncing app", "app", s)
+
+			_, err := r.syncApp(s.Name)
```
DiogoMCampos, line 102: The left-hand side variable is the synced `*argov1alpha.Application`, when there's no error. Is there anything we want to do with it?

controllers/progressiverollout_controller.go
```diff
@@ -92,11 +94,21 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
 		}
 
 		// Get the Applications to update
-		scheduledApps := scheduler.Scheduler(apps, stage)
+		scheduledApps := scheduler.Scheduler(outOfSyncApps, stage)
```
DiogoMCampos, line 97: This was using all apps, but I guess we only want to schedule the `outOfSyncApps`?

maruina: No, it's all the apps so you know how many are in sync and how many you still have to do. If it's not clear we probably need to expand the scheduler documentation.

controllers/progressiverollout_controller.go
maruina: I'm not really even sure we want to log this to be honest. This mean we tried to re sync the app while a sync was already in progress, correct? So what's the big deal? :)
I might miss a valid use case.

controllers/progressiverollout_controller.go
maruina: > Do we want to somehow signal the stage as failed

For future work in https://github.com/Skyscanner/argocd-progressive-rollout/issues/30 - as we don't do it at the moment

maruina (commented):
Just a couple of comment to answer some of your questions, I'll do a deeper review later.

config/crd/bases/deployment.skyscanner.net_progressiverollouts.yaml
```diff
@@ -1,10 +1,10 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
sledigabel, line 3: Is this the official release for v1?
Should we keep it a separate commit and release/package a specific version?

controllers/progressiverollout_controller.go
sledigabel: I guess it's to have a lambda style approach where you can do things like `r.syncApps(s.Name).Status()` or something like this in a single command.
Don't think we need to do something in particular right now.

controllers/progressiverollout_controller.go
sledigabel: What @maruina said... happy path only :-)

config/crd/bases/deployment.skyscanner.net_progressiverollouts.yaml
DiogoMCampos: Not at all. This was auto-generated by kubebuilder - I can revert the change and try to see where/why it was generated, but judging by https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/ `v1beta1` seems to have been deprecated in v1.16?

controllers/progressiverollout_controller_test.go
```diff
@@ -22,6 +25,26 @@ const (
...
+}
+
+func (c *mockArgoCDAppClientCounter) Sync(ctx context.Context, in *applicationpkg.ApplicationSyncRequest, opts ...grpc.CallOption) (*argov1alpha1.Application, error) {
+	c.m.Lock()
+	c.appsSynced = append(c.appsSynced, *in.Name)
+	defer c.m.Unlock()
```
sledigabel, line 36: as a convention, if you're deferring the unlock, it's good to keep it next to the `Lock()`

controllers/progressiverollout_controller_test.go
```diff
@@ -22,6 +25,26 @@ const (
...
+
+	return nil, nil
+}
+
+func (c *mockArgoCDAppClientCounter) GetSyncedApps() []string {
+	c.m.Lock()
+	defer c.m.Unlock()
```
sledigabel, line 43: What is the lock about here?

controllers/progressiverollout_controller.go
```diff
@@ -37,13 +37,15 @@ import (
...
 )
 
 // ProgressiveRolloutReconciler reconciles a ProgressiveRollout object
 type ProgressiveRolloutReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Log             logr.Logger
+	Scheme          *runtime.Scheme
+	ArgoCDAppClient utils.ArgoCDAppClient
```
sledigabel, line 48: does the client keep a state or is it a brand new connection every single time?

main.go
```diff
@@ -68,10 +69,19 @@ func main() {
...
+	acdAppClient := utils.GetArgoCDAppClient(c)
+
 	if err = (&controllers.ProgressiveRolloutReconciler{
-		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("ProgressiveRollout"),
-		Scheme: mgr.GetScheme(),
+		Client:          mgr.GetClient(),
+		Log:             ctrl.Log.WithName("controllers").WithName("ProgressiveRollout"),
+		Scheme:          mgr.GetScheme(),
+		ArgoCDAppClient: acdAppClient,
```
sledigabel, line 84:
```suggestion
		ArgoCDAppClient: utils.GetArgoCDAppClient(c),
```

Why not using the factory function in the reconciler definition, as you did with the other lines?

internal/utils/config.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return value, nil
+}
+
+
+/*
+ReadConfiguration returns a Configuration, reading it from environment variables if present, or from files in
+the configuration directory otherwise
+*/
+func ReadConfiguration() (*Configuration, error) {
```
sledigabel, line 51: why using pointer semantics here?
we're using value semantics everywhere else?

sledigabel: it's a minor thing, tbh, the logic of using pointer semantics makes sense so we can use `nil`. It's just against the decision made in #33.

@maruina what do you think?

config/crd/bases/deployment.skyscanner.net_progressiverollouts.yaml
sledigabel: that confused the hell out of me :-)
my bad

controllers/progressiverollout_controller.go
sledigabel: We're actually doing something funny with that one. the `sync` from Argo takes an `ApplicationRequest` that we get from the `Name`. I guess the left hand side there is meant to tell you whether the selector worked. It could be useful down the line for us.

README.md
```diff
@@ -71,6 +100,26 @@ make install
...
+### Deploying to a Kubernetes cluster
+
+To deploy the controller to a Kubernetes cluster, run:
+```
+make install
+make docker-build
+make deploy
+```
+
+In order to so, a `.env.cluster` file needs to exist in the `config/manager` folder, containing
```
maruina, line 117: I am not sure about adding a file into the kustomize directory. Is that the only way to make kustomize work?

README.md
```diff
@@ -60,6 +60,35 @@ Expect a non-functional controller and breaking changes until Milestone 2 is com
...
+```
+ARGOCD_AUTH_TOKEN=ey...
+ARGOCD_SERVER_ADDR=argocd-server
+```
+
+
+2. Files in the Config Directory.
+```
+.
+├── .prcconfig
```
maruina, line 85: Should this be `~/.prcconfig`?

config/crd/bases/deployment.skyscanner.net_progressiverollouts.yaml
maruina: If we use `apiextensions.k8s.io/v1` it will not run in our current control cluster, so please let's skip this for a little bit.

config/manager/manager.yaml
```diff
@@ -36,4 +41,8 @@ spec:
           requests:
             cpu: 100m
             memory: 20Mi
+        volumeMounts:
+            - mountPath: /.prcconfig
```
maruina, line 45: I'm not sure why we should use a `.prcconfig` in the root filesystem. Maybe `/etc/prcconfig`, or `/opt/prcconfig` or `~/.prcconfig`?

controllers/progressiverollout_controller_test.go
DiogoMCampos: There's still a read in this method (in the return statement), so we need a read lock too.

internal/utils/config.go
DiogoMCampos: Uhm - we can go with value semantics and return an empty config. As long as it also returns an error, we're able to handle it.

go.mod
```diff
@@ -7,11 +7,13 @@ require (
...
 	github.com/kr/pretty v0.2.1 // indirect
-	github.com/onsi/ginkgo v1.14.2
-	github.com/onsi/gomega v1.10.3
+	github.com/onsi/ginkgo v1.15.2
+	github.com/onsi/gomega v1.11.0
+	google.golang.org/grpc v1.33.1
 	k8s.io/api v0.20.1
 	k8s.io/apimachinery v0.20.1
 	k8s.io/client-go v11.0.1-0.20190816222228-6d55c1b1f1ca+incompatible
+	k8s.io/kubernetes v1.20.1 // indirect
```
maruina, line 16: Do you know where this is coming from?

internal/utils/config.go
maruina: Depending how we make things complicated, but it would be good to follow the same convention.

README.md
DiogoMCampos: Not very fond of it either, to be honest. But from https://kubectl.docs.kubernetes.io/faq/kustomize/: `v2.0 added a security check that prevents kustomizations from reading files outside their own directory root.` 

This meant I can't find any file that's not inside that directory - I'm open to alternatives.

config/manager/manager.yaml
DiogoMCampos: Good point, we should not. Which one do you prefer? I'm not unhappy with `~`, but maybe `/etc/` is a better place for it?

README.md
DiogoMCampos: Ah, so if I had not stopped reading at the first line, it says this: 

```Resources (including configmap and secret generators) can still be shared via the recommended best practice of placing them in a directory with their own kustomization file, and referring to this directory as a base from any kustomization that wants to use it. This encourages modularity and relocatability.```

Let me investigate that.

File: go.mod (hunk 16 above)
Comment (DiogoMCampos) [reply go.mod:16]: Tbh, no - it sure is weird because I've not installed nor updated any K8s-specific libraries. I'll get back to you on this.

File: config/manager/manager.yaml (hunk 15 above)
Comment (sledigabel) [reply config/manager/manager.yaml:45]: 👍  for `/etc/`

File: internal/utils/argocd_client.go [hunk 17]
```diff
@@ -0,0 +1,27 @@
...
+
+type ArgoCDAppClient interface {
+	Sync(ctx context.Context, in *applicationpkg.ApplicationSyncRequest, opts ...grpc.CallOption) (*v1alpha1.Application, error)
+}
+
+// GetArgoCDAppClient builds an Argo CD application client with the provided configuration
+func GetArgoCDAppClient(c *Configuration) ArgoCDAppClient {
+	acdClientOpts := argocdclient.ClientOptions{
+		ServerAddr: c.ArgoCDServerAddr,
+		Insecure:   true,
```
sledigabel, line 19: mmmmh I'm not sure about this. could we parameterise this in the Configuration object instead of leaving this insecure by default?

controllers/progressiverollout_controller_test.go
```diff
@@ -22,6 +25,26 @@ const (
 	interval = time.Millisecond * 10
 )
 
+type mockArgoCDAppClientCounter struct {
```
maruina, line 28: I think there is something I don't understand here: why is this called `Counter`? Is that because we're counting how many apps we sync?

controllers/progressiverollout_controller_unit_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"errors"
+	applicationpkg "github.com/argoproj/argo-cd/pkg/apiclient/application"
+	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
+	. "github.com/onsi/gomega"
+	"google.golang.org/grpc"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+type mockArgoCDAppClientSyncOK struct{}
```
maruina, line 14: I think there is something that I'm missing here: why we don't have a `mockArgoCDAppClient` package with its own methods and tests?
The approach would be similar in https://github.com/Skyscanner/kms-issuer/tree/main/pkg/kmsmock

controllers/suite_test.go
```diff
@@ -44,6 +47,12 @@ var k8sClient client.Client
 var testEnv *envtest.Environment
 var reconciler *ProgressiveRolloutReconciler
 
+type mockArgoCDAppClient struct{}
+
+func (*mockArgoCDAppClient) Sync(ctx context.Context, in *applicationpkg.ApplicationSyncRequest, opts ...grpc.CallOption) (*argov1alpha1.Application, error) {
```
maruina, line 52: I'm not sure I understand we we need to redefine the `mockArgoCDAppClient` and the sync method here. Can't we import it from a package?

DiogoMCampos: From what I saw, defining mocks in the test files they're used seems to be the most common practice (see https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/staging/src/k8s.io/apiserver/pkg/authentication/token/union/unionauth_test.go#L30, for an example). 

But if we prefer to have a package for the mocks, I'm happy to move it.

controllers/progressiverollout_controller_unit_test.go
DiogoMCampos: For why the mocks are in the file and not a separate package, see the other comment - I'm happy to move them if we don't want to.

There's also the fact that I want different mocks to do different things - in fact, the one in `suite_test` is more like a stub than a mock _per se_, it's there just to make sure the other tests pass by default.

controllers/progressiverollout_controller_test.go
DiogoMCampos: The idea of this mock is to mimic what in Java or Python would be `assertCalledWith()` or similar - I want to make sure that `Sync()` is called with a specific app name. 


I think I called it `Counter` because initially I was just counting number of calls - but that's wrong, we want to know that a specific app was synced, rather than how many times apps where synced. So perhaps there's a better name for it?

maruina: > The idea of this mock is to mimic what in Java or Python would be assertCalledWith() or similar - I want to make sure that Sync() is called with a specific app name.

Thanks, now it make more sense to me? Maybe `mockArgoCDAppClientWithApp`?

controllers/suite_test.go
maruina: I don't think I have good enough knowledge on the ecosystem and golang on this to be opinionated :)

I guess my point it was to be able to reuse the mock instead of redefining it multiple times? I guess when I see `mockArgoCDAppClientCounter`, `mockArgoCDAppClient` and `mockArgoCDAppClientSyncOK` they _feel_ the same thing to me with different methods? But maybe not.

@sledigabel what do you think?

DiogoMCampos: They serve different purposes in different tests, but are all implementing the same interface:

1. `mockArgoCDAppClient` is not a *mock*, it really is a generic *stub* so that we can run the other tests when we don't care about what happens in the calls to Argo CD.
2. `mockArgoCDAppClientCounter` is really just a way to `assertCalledWith` the `Sync()` method - we could use this mock as the 'general' one (in 1.) if we so wished, I opted to make the other entirely useless to make it clear we _don't_ care about its functionality.
3. `mockArgoCDAppClientSyncOK` is used in unit testing to mock that the Argo CD app client returned OK, and conversely, `SyncNotOK` returns an error. 

There's good things with this approach, imo: it's very clear what each mock does, for instance. There's also bad things: if we ever opt to use one more method from the Argo CD App Client in the code, we have to update all these mocks, which adds toil to future maintainers.

Thoughts from those with more knowldege of Golang are most welcome here 🙂

danmx: why bother creating your own mocks in the first place? Go can generate mocks for you (https://github.com/golang/mock) based on your interfaces.

DiogoMCampos: Don't see any reason not to use it, but I've never used any way of mocking in Golang, so I'm not opinionated on it.

Taking a step back then: what is the approach we want to take w.r.t. mocking in this project? If it's using `gomock`, then let's change these mocks :). Again, I'm not opinionated either way, but would rather decide on one approach and have consistency going forward.

sledigabel: yeah i was going to suggest having a /mock package would be convenient, in case we need to reuse this for other tests moving forward.
Not sure about using mock to do that, but why not if that helps in the future.

DiogoMCampos: Alright - so I did some playground stuff with gomock and I think that might be useful in the long term. Bearing in mind that the Sync is a blocker to other tasks, I've opted to simply move the mocks to a `mocks` package for the time being. We should still agree on how we're doing these things long-term.

DiogoMCampos:
Alright, I _think_ I've addressed all comments - let me know if I haven't. A few notes:
- I opted not to generate the secret with `kustomize`, as I fell Matteo is right on not wanting the `.env` file in those directories. Instead, I assume it already exists in the target cluster - for the purposes of development, the setup script generates said secret. If we are okay with this, I'll create a ticket for us to manage this secret in the future.
- I think `gomock` is actually a very good shout - although, I did run into an issue with the `Eventually`/async tests and with `gomock` expecting a set number of calls in my experimentation. For the purposes of this PR, I'm suggesting we don't use it, but I think we should give it serious consideration going forward (and revisit the mocks in here).
- As the reviews came in and the PR evolved, the configuration got a bit more complex than it was initially - at some point we should consider `viper` (or similar) instead of maintaining our own configuration manager.

controllers/progressiverollout_controller_unit_test.go
```diff
@@ -0,0 +1,35 @@
+package controllers
```
maruina, line 1: Question for the wider audience: do we want to have a separate test file for the unit tests, or can we merge those tests with `progressiverollout_controller_test.go`?

I would prefer a single test file for simplicity, but I'm open to suggestions

internal/utils/config.go
```diff
@@ -0,0 +1,92 @@
...
+func readFromEnvOrFile(paramName string) (string, error) {
+	var err error
+	value := os.Getenv(paramName)
+
+	if len(value) == 0 {
+		value, err = readFromFile(ConfigDirectory + paramName)
+		if err != nil {
+			return "", err
+		}
+	}
```
maruina, line 43:
```suggestion
	value, ok := os.LookupEnv(paramName)

	if !ok {
		value, err = readFromFile(ConfigDirectory + paramName)
		if err != nil {
			return "", err
		}
	}
```
If you use `LookupEnv` you get also a nice bool to tell you if the environment variable is set or not. What do you think?

maruina:
@DiogoMCampos I think we should also change the helm chart to exposes 
```
ArgoCDAuthToken
ArgoCDServerAddr
ArgoCDInsecure
```
in the values.yaml and create the secret. It is then up to the end-user how to pass to token to the chart.

controllers/progressiverollout_controller_unit_test.go
sledigabel: There's an argument to separate integration tests vs unit tests as they look and are different.
How is it done on other operators?

internal/utils/config.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return Configuration{}, err
+	}
+
+	acdServerAddr, err = readFromEnvOrFile(ArgoCDServerAddrKey)
+	if err != nil {
+		return Configuration{}, err
+	}
+
+	acdInsecure, err = isFlagSet(ArgoCDInsecureKey)
```
sledigabel, line 80: what's the value if the env var is not set?
Wondering if we should set defaults for those somewhere.

maruina: I don't think we can have a default for the token one, but we should definetly have one for the server address and insecure

controllers/progressiverollout_controller_unit_test.go
maruina: I have never seen anything other controller.go and controller_test.go but I am happy if anyone can findaa different example.

internal/utils/config.go
DiogoMCampos: Right now, there's no default. We fail if they're not present. 

I wouldn't be too keen on setting a default right now because we're not 100% sure on what defaults look like and there's no practical advantage _right now_. We definitely should have them in the future where we can (address, insecure). Do you think it's important to have it now? You always have to set the token anyway, it's not like you can run only on defaults.

controllers/progressiverollout_controller_unit_test.go
DiogoMCampos: In other languages (such as Python or Java), tests would be in their own folder, and I feel like I'd separate unit and other types of tests in different files too. So I felt this makes sense.

But Golang seems to be different in that tests generally live in the same package as the code under test; so perhaps a 1:1 relation between code-under-test and test-code files is ideal? If so, I can move them to the same file. I'm not opinionated on this.

internal/utils/config.go
sledigabel: Filed #45

controllers/progressiverollout_controller.go
```diff
@@ -304,3 +314,14 @@ func (r *ProgressiveRolloutReconciler) removeAnnotationFromApps(apps []argov1alp
...
 }
+
+// syncApp sends a sync request for the target app
+func (r *ProgressiveRolloutReconciler) syncApp(appName string) (*argov1alpha1.Application, error) {
+	ctx := context.Background()
+
+	syncReq := applicationpkg.ApplicationSyncRequest{
+		Name: &appName,
+	}
+
```
sledigabel, line 325: I think we have an opportunity here to validate that the app being synced is the app we passed as a parameter.
i'll raise a PR separately, it's not blocking that one.

maruina (commented):
LGTM

sledigabel (approved):
LGTM, adds a sync capability to the progressive rollout operator

sledigabel:
Are we gonna merge this?

PR #50, opened 2021-04-12
Remove required clause from secret

### LINKED ISSUE #49: Helm chart doesn't deploy without
The current `Secret.yaml` makes the data field mandatory. This breaks the following workflows:
- generating the secrets out-of-band so you don't have to store the token in git
- using secrets controller like https://github.com/external-secrets/kubernetes-external-secrets or https://github.com/bitnami-labs/sealed-secrets where they generate the secret for you.

We should remove the `required` clause from the chart.

maruina:
Fix #49


DiogoMCampos (approved):
Removes `required` clause from ArgoCD configuration secret values, and adds them to the README.

LGTM!

PR #53, opened 2021-04-13
Fix default log level

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/52.

At the moment we log in dev mode and we generate a massive amount of output. 
The approach we took here is to use https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap@v0.8.0#Options.BindFlags so we can pass some flags when starting the manager to change the logger behavior.

In particular, we can pass `--zap-devel=true` when running locally to get Development Mode defaults (encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn) in contrast to Production Mode defaults (encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error).


main.go
```diff
@@ -53,9 +53,11 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.Parse()
 
-	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	opts := zap.Options{}
+	opts.BindFlags(flag.CommandLine)
+	flag.Parse()
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
```
sledigabel, line 60: is this actually parsing all options to look for a `zap-devel`?
For some reason I can't find a doc about `UseFlagOptions`

maruina: It seems to me that is looking at all the flags and using only the zap ones

The docs are at
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap@v0.8.0#UseFlagOptions

sledigabel (approved):
LGTM,
set verbosity from a flag.

PR #55, opened 2021-04-13
Fix quoting issues

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/51


maruina:
Looking at https://github.com/Skyscanner/argocd-progressive-rollout/runs/2331651220?check_suite_focus=true it seems the installation is successful.

sledigabel (approved):
Removing the quotes before encoding, doesn't work well with the bool type

PR #57, opened 2021-04-13
Fix ginkgo async testing

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/56


maruina:
After having a chat in the #kubebuilder channel, it seems there are some changes needed in the controller-runtime itself before making this change. See https://kubernetes.slack.com/archives/CAR30FCJZ/p1618306852445800 for more context.

Closing for the time being.

PR #58, opened 2021-04-13
Boilerplate for setting up remote debugging

nebojsa-prodana:
Allows us to create a debug version of the image that runs delve debugging server that we could attach to. Useful for debugging both within the local kind cluster and potentially we could consider deploying a debug image as well when we need to debug something in production.

Currently missing:
* A way to kustomize adding ports to the manager container to allow connecting to delve
* Creation of socat proxy for exposing port 2345 on the local network
* Readme instructions
* A make target/script to setup the boilerplating for debugging
* Configuration for IDE's other than vscode for attaching to the remote debugger.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,55 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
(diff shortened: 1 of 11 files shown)


maruina:
Close because addressed in #97

PR #59, opened 2021-04-14
Use golang tools best practice

maruina:
This PR use the recommended approach described in https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

`make tools` provide an easy way to install the required tools to interact with this project with a version pinned in `go.mod`

This fix the issue where people can have a local version of `ginkgo` or `controller-gen` that is different from the one used in `go.mod`. For example at one point I realized I was:
- running a local ginkgo v1.10.0
- using ginkgo v1.15.2 for writing tests as per go.mod
- using gingko v1.16.1 when doing the github action. See for example https://github.com/Skyscanner/argocd-progressive-rollout/runs/2333896133?check_suite_focus=true#step:6:11

The controller-gen version is coming from https://github.com/kubernetes-sigs/kubebuilder/blob/v2.3.2/testdata/project-v2/Makefile#L74

You can see this approach in action here: https://github.com/Skyscanner/argocd-progressive-rollout/runs/2342308815?check_suite_focus=true


sledigabel (approved):
LGTM, hopefully this will fix my problem :-)

PR #60, opened 2021-04-15
Add stages status

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/8


api/v1alpha1/stage_status_types.go
```diff
@@ -0,0 +1,87 @@
...
+	FinishedAt metav1.Time      `json:"finishedAt,omitempty"`
+}
+
+// GetStageStatus returns a pointer to the Status.Stages slice
+func (in *ProgressiveRollout) GetStageStatus() *[]StageStatus {
+	return &in.Status.Stages
+}
+
+// NewStageStatus adds a new StageStatus
+func (in *ProgressiveRollout) NewStageStatus(n, m string, p StageStatusPhase, t, s, r, f, c []string) StageStatus {
```
nebojsa-prodana, line 38: why is completed for instance a list of strings? I am still getting used to golang conventions, but it is quite hard to navigate all these single letter variables, could we make them more descriptive?

api/v1alpha1/stage_status_types.go
```diff
@@ -0,0 +1,87 @@
...
+		FinishedAt: metav1.Time{},
+	}
+}
+
+// SetStageStatus sets the corresponding StageStatus in stageStatus to newStatus
+func SetStageStatus(stageStatus *[]StageStatus, newStatus StageStatus) {
+	existingStatus := FindStageStatus(*stageStatus, newStatus.Name)
+
+	if existingStatus == nil {
+		*stageStatus = append(*stageStatus, newStatus)
```
nebojsa-prodana, line 58: Am I reading this right: Can a single stage have multiple statuses set in a single pass of the controller logic?

api/v1alpha1/stage_status_types.go
maruina: Because of the output we discussed on https://github.com/Skyscanner/argocd-progressive-rollout/issues/8? If not a list of string, what would you use?

api/v1alpha1/stage_status_types.go
maruina: I don't think so, where do you see that?

api/v1alpha1/stage_status_types.go
maruina: You are correct, we should use more descriptive parameters. I'm going to fix it here and fix as well in https://github.com/Skyscanner/argocd-progressive-rollout/blob/main/api/v1alpha1/progressiverollout_types.go#L83 that is the only other function that uses single letter names.

api/v1alpha1/stage_status_types.go
sledigabel: that confuses me too; `append` makes it look like `stageStatus` is an array, and it is.
do you mean `stageList` rather than `stageStatus`?

maruina: Ah, I see it know, thanks.
Yes, it should probably be `stageStatusList *[]StageStatus` or just `stageList`

api/v1alpha1/progressiverollout_types.go
```diff
@@ -73,13 +73,15 @@ type ProgressiveRolloutStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	Stages     []StageStatus      `json:"stages,omitempty"`
```
sledigabel, line 76: we use pointer semantics practically everywhere in this; wondering if it would be worth it to make it a list of pointers.
```suggestion
	Stages     []*StageStatus      `json:"stages,omitempty"`
```

api/v1alpha1/stage_status_types.go
sledigabel: you mean `[]*StateStatus` :)

api/v1alpha1/progressiverollout_types.go
maruina: The reasons why I suggested value semantics are:
- we use value semantics everywhere, and I wanted to keep consistency
- for `metav1.Condition`, when I looked at some operators like https://github.com/fluxcd/kustomize-controller/blob/main/api/v1beta1/kustomization_types.go#L212 or https://github.com/fluxcd/flagger/blob/main/pkg/apis/flagger/v1beta1/status.go#L85 or https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L544 they all seems to use value semantic
- there is an official package to deals with Conditions at https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta but those function requires a pointer to the slice. This is why we have the `GetStatusConditions()` method at https://github.com/Skyscanner/argocd-progressive-rollout/blob/main/api/v1alpha1/progressiverollout_types.go#L79

So to me it makes sense to keep the same semantic, and possibly the same names for our functions, to interact with StageStatus. This is why we have a `GetStageStatus()` method for example.

I'm happy to discuss if we're not doing a great job at the using the same approach as the `api/meta` package, or if we would be better with a different approach, but would feel weird to have values for every slice _except_ for this one

sledigabel: this is a good rationale;
let's make sure we're not getting confused with this and that the "true" value resides with this one.

api/v1alpha1/stage_status_types.go
```diff
@@ -0,0 +1,79 @@
...
+type StageStatus struct {
+	Name       string           `json:"name"`
+	Phase      StageStatusPhase `json:"stage,omitempty"`
+	Message    string           `json:"message,omitempty"`
+	StartedAt  *metav1.Time     `json:"startedAt,omitempty"`
+	FinishedAt *metav1.Time     `json:"finishedAt,omitempty"`
+}
+
+// GetStageStatus returns a pointer to the Status.Stages slice
+func (in *ProgressiveRollout) GetStageStatus() *[]StageStatus {
```
sledigabel, line 27: this should be a slice, not a pointer to a slice

api/v1alpha1/stage_status_types.go
```diff
@@ -0,0 +1,79 @@
...
+	return &in.Status.Stages
+}
+
+// NewStageStatus adds a new StageStatus
+func (in *ProgressiveRollout) NewStageStatus(name, message string, phase StageStatusPhase) StageStatus {
```
sledigabel, line 32: i don't really see why this needs to be a method of `ProgressiveRollout`?

api/v1alpha1/stage_status_types.go
sledigabel: (generally speaking, there should not be any pointer to slices, it's already a pointer)

api/v1alpha1/stage_status_types.go
```diff
@@ -0,0 +1,79 @@
...
+		Message:    message,
+		StartedAt:  nil,
+		FinishedAt: nil,
+	}
+}
+
+// SetStageStatus sets the corresponding StageStatus in stageStatus to newStatus
+// - If a stage doesn't exist, it will be added to StageStatus slice
+// - If a stage already exists it will be updated
+func SetStageStatus(stageStatus *[]StageStatus, newStatus StageStatus, updateTime *metav1.Time) {
```
sledigabel, line 45: this should be a slice, not a pointer to a slice
```suggestion
func SetStageStatus(stageStatus []StageStatus, newStatus StageStatus, updateTime *metav1.Time) {
```

sledigabel: on second read, i see what you've done here. I'm not sure this function signature is really the best here.
unlike `NewStageStatus`, that one should prob be a method of `ProgressiveRollout`.

Something like:
```suggestion
func (p *ProgressiveRollout) SetStageStatus(newStatus StageStatus, updateTime *metav1.Time) {
```

The name `SetStageStatus` i struggled with a little bit... , but that's secondary.

sledigabel: on third read :-) i'd prob suggest to create two functions instead, like `AddStageStatus` and `UpdateStageStatus`, and handle the `FindStageStatus` in the `updateStageStatus` at line 351

api/v1alpha1/progressiverollout_types.go
```diff
@@ -100,6 +102,34 @@ func (in *ProgressiveRollout) Owns(owners []metav1.OwnerReference) bool {
 	return false
 }
 
+// SetStageStatus sets the corresponding StageStatus in stageStatus to newStatus
+// - If a stage doesn't exist, it will be added to StageStatus slice
+// - If a stage already exists it will be updated
+func (in *ProgressiveRollout) SetStageStatus(newStatus StageStatus, updateTime *metav1.Time) {
```
sledigabel, line 108: we briefly mentioned it earlier, but is there a reason behind the pointer semantics of  `metav1.Time`?

maruina:
To be able to have it as `nil` so you can have only the `startedAt` and
omit the `completedAt`

sledigabel:
> To be able to have it as `nil` so you can have only the `startedAt` and omit the `completedAt`

Ha, yes :-)

api/v1alpha1/progressiverollout_types.go
```diff
@@ -100,6 +102,34 @@ func (in *ProgressiveRollout) Owns(owners []metav1.OwnerReference) bool {
 	return false
 }
 
+// SetStageStatus sets the corresponding StageStatus in stageStatus to newStatus
+// - If a stage doesn't exist, it will be added to StageStatus slice
+// - If a stage already exists it will be updated
+func (in *ProgressiveRollout) SetStageStatus(newStatus StageStatus, updateTime metav1.Time) {
```
sledigabel, line 108: so if we're going to leave a pointer for Time, in this instance we should have a pointer parameter too;
that will avoid having to copy `updateTime` as parameter, and then send the pointer back which means the struct will leave the stack and go to the heap.
That will cause some unnecessary garbage collection

api/v1alpha1/progressiverollout_types.go
DiogoMCampos: On this, we kept the time as a pointer in the struct for the reasons Matteo mentioned, but changed it in this method's signature - I can change it back if we go with pointer semantics here anyway.

api/v1alpha1/progressiverollout_types.go
DiogoMCampos: Alright - I'll change it back - I just asked this in another comment 😛

controllers/progressiverollout_controller.go
```diff
@@ -71,67 +72,117 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
 		log = r.Log.WithValues("stage", stage.Name)
 
 		// Get the clusters to update
-		clusters, err := r.getClustersFromSelector(stage.Targets.Clusters.Selector)
-		if err != nil {
-			log.Error(err, "unable to fetch clusters")
-			return ctrl.Result{}, err
+		clusters, cErr := r.getClustersFromSelector(stage.Targets.Clusters.Selector)
```
DiogoMCampos, line 75: I'm not particularly fond of these `cErr`, `aErr`, etc. I'd be tempted to keep them in for now and refactor later, as I feel the reconciliation method may need significant refactoring as it's currently very long and difficult to follow.

controllers/progressiverollout_controller.go
```diff
@@ -71,67 +72,119 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+
+			failedMessage := fmt.Sprintf("%s stage failed", stage.Name)
+			failed := pr.NewStatusCondition(deploymentskyscannernetv1alpha1.CompletedCondition, metav1.ConditionTrue, deploymentskyscannernetv1alpha1.StagesFailedReason, failedMessage)
+			apimeta.SetStatusCondition(pr.GetStatusConditions(), failed)
+			if err := r.patchStatus(ctx, &pr); err != nil {
+				r.Log.Error(err, "failed to update object status")
+				return ctrl.Result{}, err
+			}
+			r.Log.Info("rollout failed")
+			// We can set Requeue: true once we have a timeout in place
```
DiogoMCampos, line 151: Currently, stage failed fails the rollout and doesn't retry - this doesn't leave much space for 'eventual consistency'. However, until we have a `progressDeadlineSeconds` in place, I'd keep it this way to avoid infinite loops.

controllers/progressiverollout_controller.go
maruina: I agree, the problem is that if you re-use `err` you are overwriting the original value, so I'm happy to get any suggestion on that. 
I agree with you that we need some (heavy) refactor.

DiogoMCampos:
Alright - I'm opening the PR for review. A few notes:
- I've completely replaced the existing test for a single stage app as I don't think there's any scenario that test covered that the new one (with two stages) doesn't cover.
- The helper methods in the tests are not ideal - I like the `createClusters` but not necessarily the `createApplications -, so I welcome any suggestions, or alternatively I'll file an issue so we look into them if we have capacity in the future.
- I opted to add `PhaseUnknown` as a stage phase: the rationale behind it is that `PhaseSucceeded` is only an alternative to `PhaseFailed` and `PhaseProgressing` if the conditions for these two match *all* possible non-successful scenarios. With `PhaseUnknown`, we declare what `PhaseSucceeded` is explicitly and let all other scenarios fall through to `PhaseUnknown`. We can hopefully iterate on this by understanding what situations get to that phase in production.

This is not perfect by any stretch, but I feel it's OK-ish to keep us moving.

As said above, we probably should look into refactoring the reconciliation loop - potentially after we have a clearer idea of what it looks like with the detailed stages info from #64.

api/v1alpha1/stage_status_types.go
DiogoMCampos: Did the first two suggested changes - but opted not to make the change to the method just now as it provides a simple single interface to make these change. Perhaps we can rename it to make it clearer?

maruina:
@DiogoMCampos what do you think about moving the scheduler change on a separate PR?

maruina (commented):
A couple of comments, mostly about splitting a couple of pieces of this PR into separate PRs so we can try to reduce this beast :)

api/v1alpha1/stage_status_types.go
```diff
@@ -0,0 +1,52 @@
...
+type StageStatus struct {
+	Name       string           `json:"name"`
+	Phase      StageStatusPhase `json:"stage,omitempty"`
+	Message    string           `json:"message,omitempty"`
+	StartedAt  *metav1.Time     `json:"startedAt,omitempty"`
+	FinishedAt *metav1.Time     `json:"finishedAt,omitempty"`
+}
+
+// GetStageStatus returns the Status.Stages slice
+func (in *ProgressiveRollout) GetStageStatus() []StageStatus {
```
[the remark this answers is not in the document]
DiogoMCampos: Completely equivalent, yea. I actually meant to ask you on Friday, but didn't want to bother you 😛 I'll do that 👍

controllers/progressiverollout_controller.go
```diff
@@ -325,3 +396,13 @@ func (r *ProgressiveRolloutReconciler) syncApp(appName string) (*argov1alpha1.Ap
 
 	return r.ArgoCDAppClient.Sync(ctx, &syncReq)
 }
+
+// patchStatus patches the progressive rollout object status
+func (r *ProgressiveRolloutReconciler) patchStatus(ctx context.Context, pr *deploymentskyscannernetv1alpha1.ProgressiveRollout) error {
```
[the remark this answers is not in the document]
DiogoMCampos: Ah, maybe we do yea - I may have misunderstood https://github.com/Skyscanner/argocd-progressive-rollout/issues/33. Let me go and rework this, thanks!

controllers/progressiverollout_controller.go
sledigabel: yeah it's common practice to keep and overwrite `err` every time.
do we need to keep `cErr`, `aErr` and `rErr` for a particular reason?

sledigabel: https://play.golang.org/p/JniiUIquyoL
`err` doesn't leave the block where it's defined, if it's assigned in the `if` statement.

DiogoMCampos: I _think_ it's addressed by https://github.com/Skyscanner/argocd-progressive-rollout/pull/60/commits/f8328b614002fdc58bb334da165ce65f35bc947e.

sledigabel (approved):
LGTM 👍  implements basic stage status

PR #61, opened 2021-04-16
Allow to reference an existing secret

maruina:
We should allow users to reference an existing secret for fetching the operator configuration.

This option allows users to generate the secret in the way they want - using a controller, manually, etc. - without having to store the tokens as clear text in git.

### DIFF
--- a/charts/argocd-progressive-rollout/README.md
+++ b/charts/argocd-progressive-rollout/README.md
@@ -28,15 +28,16 @@ Users of Helm v2 should use `--set installCRDs=true` to install the CRDs.
 | affinity | object | `{}` |  |
 | args.enableLeaderElection | bool | `false` |  |
 | args.metricsAddr | string | `":8080"` |  |
-| config.argoCDAuthToken | string | `"example-token"` |  |
-| config.argoCDInsecure | string | `"true"` |  |
-| config.argoCDServerAddr | string | `"argocd-server"` |  |
-| configSecret.annotations | object | `{}` |  |
-| configSecret.name | string | `""` |  |
+| config | object | `{"argoCDAuthToken":"example-token","argoCDInsecure":"true","argoCDServerAddr":"argocd-server"}` | Config options |
+| config.argoCDAuthToken | string | `"example-token"` | ArgoCD token |
+| config.argoCDInsecure | string | `"true"` | Allow insecure connection with ArgoCD server |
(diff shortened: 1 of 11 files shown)


Smirl (approved):
LGTM.

The secret that is created has a hard coded name. Shouldn't be a problem unless you deploy 2 of the controller to the same ns which isn't likely.

The key of the secret might change, at the moment it has to be `ARGOCD_AUTH_TOKEN`, etc. Again, not the end of the world for now as they are unique enough keys.

maruina:
@Smirl I pushed an update after you comment:
- The secret name is now dynamic.
- By default, we create the secret instead of using the external one.
- There 2 tests: one for the generated secret and one for the external one.
- In order to have an external secret in the kind cluster before running the tests, I had to install it and run the tests in a predictable namespace.

charts/argocd-progressive-rollout/README.md
```diff
@@ -28,9 +28,11 @@ Users of Helm v2 should use `--set installCRDs=true` to install the CRDs.
...
 | args.enableLeaderElection | bool | `false` |  |
 | args.metricsAddr | string | `":8080"` |  |
-| configSecret.argoCDAuthToken | string | `"example-token"` |  |
-| configSecret.argoCDInsecure | string | `"true"` |  |
-| configSecret.argoCDServerAddr | string | `"argocd-server"` |  |
+| config.argoCDAuthToken | string | `"example-token"` |  |
+| config.argoCDInsecure | string | `"true"` |  |
+| config.argoCDServerAddr | string | `"argocd-server"` |  |
+| configSecret.annotations | object | `{}` |  |
+| configSecret.name | string | `""` |  |
```
sledigabel, line 35: Should we add some documentation here as to what this does? either here or in the main readme?
just mentioning somewhere that if `configSecret.name` is not specified, the helm chart will generate a secret and use it.

sledigabel (dismissed):
Minor comment about the doc, but overall LGTM.

Adds the ability to configure and maintain a secret separately from the helm chart.

charts/argocd-progressive-rollout/README.md
Smirl: This parameter is often called `existingSecret` in many other charts that I have used.

maruina: I pushed a change to add a description

sledigabel (approved):
LGTM

PR #68, opened 2021-05-03
Fix flaky test

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/62

Because the k8s API server may not have finished creating the object after our `Create()` call from earlier, we use `Eventually()` testing function instead of `Expect()` to give the API server an opportunity to finish creating the object.

We also remove unnecessary `Create()` since when we call `requestsForApplicationChange` and `requestsForSecretChange` we pass the application and secret as parameter and they don't need to be on the API server.

### DIFF
--- a/controllers/progressiverollout_controller_test.go
+++ b/controllers/progressiverollout_controller_test.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/uuid"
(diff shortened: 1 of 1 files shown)


DiogoMCampos (approved):
Improves a flaky test by allowing for eventual consistency instead of assuming immediacy from the call to the K8s API. LGTM!

sledigabel:
We actually used it before iirc.
Good find @maruina !

sledigabel (approved):
LGTM, adds the Eventually() stanza to cater for eventual consistency

PR #72, opened 2021-05-10
Scheduler and Util changes for stages status

DiogoMCampos:
Scheduler and utilities changes from #60 to make it easier to review the PRs.

### DIFF
--- a/controllers/progressiverollout_controller.go
+++ b/controllers/progressiverollout_controller.go
@@ -19,6 +19,8 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
+
 	deploymentskyscannernetv1alpha1 "github.com/Skyscanner/argocd-progressive-rollout/api/v1alpha1"
 	"github.com/Skyscanner/argocd-progressive-rollout/internal/scheduler"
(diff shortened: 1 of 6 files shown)


controllers/progressiverollout_controller.go
```diff
@@ -109,21 +110,13 @@ func (r *ProgressiveRolloutReconciler) Reconcile(ctx context.Context, req ctrl.R
...
-		if scheduler.IsStageComplete(apps, stage) {
-			// TODO: update status
-			r.Log.Info("stage completed")
-		} else {
-			// TODO: update status
-			r.Log.Info("stage in progress")
-			// Stage in progress, we reconcile again until the stage is completed or failed
-			return ctrl.Result{Requeue: true}, nil
-		}
+		r.Log.Info("stage completed")
```
DiogoMCampos, line 119: This is an 'odd change' - it's perfectly backwards compatible with what's in `main` as `IsStageComplete(apps, stage)` always returned true. But ensures backwards compatibility for tests as `IsStageComplete` has an actual body now and tests would fail otherwise.

It'll be promptly changed again in #60.

maruina (approved):

PR #74, opened 2021-05-10
Use client for asserting against the live API server

### LINKED ISSUE #73: Fix k8sClient in test
[According to the latest kubebuilder book about writing test](https://book.kubebuilder.io/cronjob-tutorial/writing-tests.html)

```
Note that we set up both a “live” k8s client, separate from the manager. This is because when making assertions in tests, you generally want to assert against the live state of the API server. If you used the client from the manager (k8sManager.GetClient), you’d end up asserting against the contents of the cache instead, which is slower and can introduce flakiness into your tests. We could use the manager’s APIReader to accomplish the same thing, but that would leave us with two clients in our test assertions and setup (one for reading, one for writing), and it’d be easy to make mistakes.
```

We should change https://github.com/Skyscanner/argocd-progressive-rollout/blob/main/controllers/suite_test.go#L100 to use a different client.

maruina:
Fix #73

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -17,6 +17,11 @@
 package controllers
 
 import (
+	"math/rand"
+	"path/filepath"
+	"testing"
+	"time"
+
 	deploymentskyscannernetv1alpha1 "github.com/Skyscanner/argocd-progressive-rollout/api/v1alpha1"
 	"github.com/Skyscanner/argocd-progressive-rollout/mocks"
 	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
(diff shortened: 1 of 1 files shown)


Smirl (approved):
LGTM

However, I noticed there are a few places that use the reconsiler's k8sClient in assertions. For example I see a lot of `reconciler.requestsForApplicationChange(ownedApp)`.

https://github.com/Skyscanner/argocd-progressive-rollout/blob/main/controllers/progressiverollout_controller_test.go#L97-L100

maruina:
Ah, I don't think there is a way around that.

The alternative is making `reconciler.requestsForApplicationChange(ownedApp)` a function instead of a method and passing the client explicitly. I _think_ the `Eventually` should cover that, we will revisit if it's not the case.

sledigabel:
I assumed this was linked somehow.
i'll propose a way and feel free to destroy it or help me refactor

Issue #81, opened 2021-05-12
Race condition during testing


sledigabel:
Found this during CI from #78

https://github.com/Skyscanner/argocd-progressive-rollout/pull/78/checks?check_run_id=2563926011#step:6:115
The test `[It] should forward events for owned applications` is failing because (I think) the namespace is global but reset before each test, which could lead to a race condition where the namespace is changed during the reconcile loop and ends up being inconsistent.

The namespace is set here:
https://github.com/Skyscanner/argocd-progressive-rollout/blob/cb03eef8d637218632005b4266b5a4e47a37b55d/controllers/progressiverollout_controller_test.go#L48-L52

and the test is performed here:
https://github.com/Skyscanner/argocd-progressive-rollout/blob/cb03eef8d637218632005b4266b5a4e47a37b55d/controllers/progressiverollout_controller_test.go#L86-L108

maruina:
I think this is the downside of using the controller client as @Smirl pointed out in https://github.com/Skyscanner/argocd-progressive-rollout/issues/73

Unfortunately I don't know how to make `requestsForApplicationChange` a function because of https://github.com/Skyscanner/argocd-progressive-rollout/pull/74#issuecomment-838394001

sledigabel:
I assumed this was linked somehow.
i'll propose a way and feel free to destroy it or help me refactor

sledigabel:
Closed via #83

PR #83, opened 2021-05-12
Fixing ginkgo Before race condition

sledigabel:
The BeforeEach from the context relies on the BeforeEach on the describe
to be completed first;
We had an issue (#81) where this wasn't the case and resulting in
inconsistent testing (flaky).

Using JustBefore will make sure the Context setup will run after the
Describe.

https://onsi.github.io/ginkgo/#separating-creation-and-configuration-justbeforeeach

### DIFF
--- a/controllers/progressiverollout_controller_test.go
+++ b/controllers/progressiverollout_controller_test.go
@@ -63,7 +63,7 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 
 	Describe("requestsForApplicationChange function", func() {
 
-		BeforeEach(func() {
+		JustBeforeEach(func() {
 			ownerPR = &deploymentskyscannernetv1alpha1.ProgressiveRollout{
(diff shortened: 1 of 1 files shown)


maruina:
Nice one, good spot! I wonder if we need the 
```
AfterEach(func() {
			Expect(k8sClient.Delete(ctx, ownerPR)).To(Succeed())
		})
```
at all. What do you think?

sledigabel:
> Nice one, good spot! I wonder if we need the
> 
> ```
> AfterEach(func() {
> 			Expect(k8sClient.Delete(ctx, ownerPR)).To(Succeed())
> 		})
> ```
> 
> at all. What do you think?

I agree, I think it's optional now, but if it doesn't hurt should we keep it? it's an extra test confirming the deletion works, still useful?
happy either way.

sledigabel:
It's good practice to clean up after the test anyway, i think we can keep it personally

maruina (approved):
LGTM

PR #84, opened 2021-05-14
Rename to `applicationset-progressive-sync` controller

maruina:
Fix https://github.com/Skyscanner/argocd-progressive-rollout/issues/66

Notes:

- Replace `deployment.skyscanner.net` with `argoproj.skyscanner.net` API Group. This is also now consistent with kms-issuer (see https://github.com/Skyscanner/kms-issuer/blob/main/api/v1alpha1/groupversion_info.go#L29)
- Replace `ProgressiveRollout` with `ProgressiveSync`
- Replace `argocd-progressive-rollout` with `applicationset-progressive-sync`


PROJECT
```diff
@@ -1,7 +1,7 @@
-domain: deployment.skyscanner.net
-repo: github.com/Skyscanner/argocd-progressive-rollout
+domain: argoproj.skyscanner.net
```
sledigabel, line 1: why this one?

maruina: I added a note in the PR description. This is to:
- have consistency between our open source controllers. In the kms-issuer we have `cert-manager.skyscanner.net` as API group, here I just used the AppSet api group.
- if we ever build another controller targeting something in the argo ecosystem we can have the same API Group with a different CRD.

Note that if the project gets incubated we will probably need to change this again :)

sledigabel: no worries, apologies I missed your note on this 👍

sledigabel (approved):
LGTM, removing all refs to argocd-progressive-rollout and refactoring to applicationset-progressive-sync

PR #85, opened 2021-05-18
Ensure objects are created in a consistent way without pointers

### LINKED ISSUE #71: Inconsistent object declaration
We are not consistent when we declare a new object.

Sometimes we do something like
```
pr := deploymentskyscannernetv1alpha1.ProgressiveRollout{}

```
and sometimes we do
```
pr := &deploymentskyscannernetv1alpha1.ProgressiveRollout{}
```

We agreed on using the first format because then it's very explicit when you need to pass a reference to it. For example
```
if err := r.Get(ctx, req.NamespacedName, &pr); err != nil {...}
```
is more clear then
```
if err := r.Get(ctx, req.NamespacedName, pr); err != nil {...}
```

Smirl:
# Summary

- fixes #71 
- ensure that all objects/structs are created as their type, then referenced when used as per issue

### DIFF
--- a/controllers/progressivesync_controller_test.go
+++ b/controllers/progressivesync_controller_test.go
@@ -83,7 +83,7 @@ var _ = Describe("ProgressiveSync Controller", func() {
 
 		It("should forward events for owned applications", func() {
 			By("creating an owned application")
-			ownedApp := &argov1alpha1.Application{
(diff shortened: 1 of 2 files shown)


maruina:
Did you have a look at the controller as well?

maruina (approved):
LGTM

PR #86, opened 2021-05-18
Fix potential race condition and remove ambiguity with BeforeEach

sledigabel:
This PR attempts to fix several race conditions that manifest in different ways:
#83, #81

I suspect this is coming from the use of `BeforeEach`, resetting the `namespace` name but being used later in the test.
It is a global scope, so by making it local I think we'll remove the issue.

### DIFF
--- a/controllers/progressivesync_controller_test.go
+++ b/controllers/progressivesync_controller_test.go
@@ -516,8 +516,8 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 
 	Describe("sync application", func() {
 		It("should send a request to sync an application", func() {
-			mockedArgoCDAppClient := &mocks.MockArgoCDAppClientCalledWith{}
-			reconciler.ArgoCDAppClient = mockedArgoCDAppClient
+			mockedArgoCDAppClient := mocks.MockArgoCDAppClientCalledWith{}
+			reconciler.ArgoCDAppClient = &mockedArgoCDAppClient
 			testAppName := "single-stage-app"
 
 			By("creating an ArgoCD cluster")
--- a/controllers/progressivesync_controller_test.go
+++ b/controllers/progressivesync_controller_test.go
@@ -37,12 +37,11 @@ var (
 func createRandomNamespace() (string, *corev1.Namespace) {
 	namespace := "progressiverollout-test-" + randStringNumber(5)
 
-	ns := &corev1.Namespace{
+	ns := corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{Name: namespace},
 	}
-	err := k8sClient.Create(ctx, ns)
-	Expect(err).To(BeNil(), "failed to create test namespace")
-	return namespace, ns
+	Expect(k8sClient.Create(ctx, &ns)).To(Succeed())
+	return namespace, &ns
 }
 
 func createOwnerPR(ns string, owner string) *syncv1alpha1.ProgressiveSync {
@@ -97,8 +96,7 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 				Name:      "owner-pr",
 			}))
 			Expect(k8sClient.Delete(ctx, ownerPR)).To(Succeed())
-			err := k8sClient.Delete(ctx, ns)
-			Expect(err).To(BeNil(), "failed to delete test namespace")
+			Expect(k8sClient.Delete(ctx, ns)).To(Succeed())
 		})
 
 		It("should filter out events for non-owned applications", func() {
@@ -125,8 +123,7 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 				return len(requests)
 			}).Should(Equal(0))
 			Expect(k8sClient.Delete(ctx, ownerPR)).To(Succeed())
-			err := k8sClient.Delete(ctx, ns)
-			Expect(err).To(BeNil(), "failed to delete test namespace")
+			Expect(k8sClient.Delete(ctx, ns)).To(Succeed())
 		})
 	})
 
@@ -168,8 +165,7 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 				return len(requests)
 			}).Should(Equal(1))
 			Expect(k8sClient.Delete(ctx, ownerPR)).To(Succeed())
-			err := k8sClient.Delete(ctx, ns)
-			Expect(err).To(BeNil(), "failed to delete test namespace")
+			Expect(k8sClient.Delete(ctx, ns)).To(Succeed())
 		})
 
 		It("should not forward an event for a generic secret", func() {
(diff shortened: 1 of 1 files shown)


controllers/progressivesync_controller_test.go
```diff
@@ -29,60 +29,50 @@ const (
 	interval = time.Millisecond * 10
 )
 
-var _ = Describe("ProgressiveSync Controller", func() {
+var (
+	appSetAPIRef = utils.AppSetAPIGroup
```
sledigabel, line 33: I tried to make this a constant but it's a pointer already so cannot be made `const`

controllers/progressivesync_controller_test.go
```diff
@@ -553,7 +564,7 @@ var _ = Describe("ProgressiveSync Controller", func() {
 			Expect(k8sClient.Create(ctx, &singleStageApp)).To(Succeed())
 
 			By("creating a progressive sync")
-			singleStagePR = &syncv1alpha1.ProgressiveSync{
+			singleStagePR := syncv1alpha1.ProgressiveSync{
```
sledigabel, line 567: This is a follow up on #71

controllers/progressivesync_controller_test.go
```diff
@@ -29,60 +29,50 @@ const (
...
-	var (
-		ctx                     context.Context
-		namespace, appSetAPIRef string
-		ns                      *corev1.Namespace
-		ownerPR, singleStagePR  *syncv1alpha1.ProgressiveSync
-	)
+func createRandomNamespace() (string, *corev1.Namespace) {
+	namespace := "progressiverollout-test-" + randStringNumber(5)
+
+	ns := &corev1.Namespace{
```
maruina, line 40:
```suggestion
	ns := corev1.Namespace{
```
as per the convention introduced in https://github.com/Skyscanner/applicationset-progressive-sync/pull/85

controllers/progressivesync_controller_test.go
```diff
@@ -106,10 +96,16 @@ var _ = Describe("ProgressiveSync Controller", func() {
 				Namespace: namespace,
 				Name:      "owner-pr",
 			}))
+			Expect(k8sClient.Delete(ctx, ownerPR)).To(Succeed())
+			err := k8sClient.Delete(ctx, ns)
+			Expect(err).To(BeNil(), "failed to delete test namespace")
```
maruina, line 101:
```suggestion
			Expect(k8sClient.Delete(ctx, ns)).To(Succeed())
```

controllers/progressivesync_controller_test.go
```diff
@@ -29,60 +29,50 @@ const (
...
-		ownerPR, singleStagePR  *syncv1alpha1.ProgressiveSync
-	)
+func createRandomNamespace() (string, *corev1.Namespace) {
+	namespace := "progressiverollout-test-" + randStringNumber(5)
+
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{Name: namespace},
+	}
+	err := k8sClient.Create(ctx, ns)
+	Expect(err).To(BeNil(), "failed to create test namespace")
```
maruina, line 44:
```suggestion
	Expect(k8sClient.Create(ctx, ns)).To(Succeed())
```

maruina (commented):
To be honest, I'm not sure about removing the `BeforeEach` block. Looking at https://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach it seems its use case is to share common setup across tests.

Here we're removing the common setup and we're using `namespace, ns := createRandomNamespace()` in every block. It seems a good candidate to be part of a `BeforeEach`.

If we want to make sure a namespace is deleted before starting another test, maybe we can add a condition like https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller_test.go#L406 in `AfterEach`?

I think we can still drop the `JustBeforeEach` and be explicit about creating the ProgressiveSync object.

controllers/progressivesync_controller_test.go
sledigabel: good catch

sledigabel:
The alternative is to make the namespace constant for all tests.
I won't mind, but I thought testing things in separate namespaces was quite nice

maruina:
> The problem is the fact that the namespace and ns changes, and that they are reused later to validate the test.
From the docs and examples I've seen, the BeforeEach is just used to reset some data or variable with a constant (we use a randomized item).

When I look at the book example in https://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach I struggle to see the difference between having a static book and something that might be random.
The namespace defined at https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller_test.go#L48 should change every `Describe` block but should be the same within the same block.

sledigabel:
> > The problem is the fact that the namespace and ns changes, and that they are reused later to validate the test.
> > From the docs and examples I've seen, the BeforeEach is just used to reset some data or variable with a constant (we use a randomized item).
> 
> When I look at the book example in https://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach I struggle to see the difference between having a static book and something that might be random.
> The namespace defined at https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller_test.go#L48 should change every `Describe` block but should be the same within the same block.

I agree with you, but in practice we've seen this not working with namespaces being inconsistent.
Unfortunately the logs have disappeared from github action, but we could see clearly that a new name had been generated between the time the namespace and the ownerpr had been created, and the time the `Expect` was being tested.

In all examples I could find, BeforeEach sets the data properly, no problem there, but our issue is that we reuse it a bit further in the test as validation and I think that's what is causing the problem.
https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller_test.go#L106

It loses elegance, but at least having the data scoped locally will not create dependencies between the tests.

sledigabel:
After chatting with @maruina , we decided to put the namespace creation back in `BeforeEach` as it is a nice feature to declutter the code.
The `ownerPR` is now delegated to the `It` section, and *should* help fix the various race conditions.

Should they happen again, we can simply move the namespace creation logic into each test.

controllers/progressivesync_controller_test.go
```diff
@@ -518,8 +516,8 @@ var _ = Describe("ProgressiveSync Controller", func() {
 
 	Describe("sync application", func() {
 		It("should send a request to sync an application", func() {
-			mockedArgoCDAppClient := mocks.MockArgoCDAppClientCalledWith{}
-			reconciler.ArgoCDAppClient = &mockedArgoCDAppClient
+			mockedArgoCDAppClient := &mocks.MockArgoCDAppClientCalledWith{}
+			reconciler.ArgoCDAppClient = mockedArgoCDAppClient
```
maruina, line 520:
```suggestion
			mockedArgoCDAppClient := mocks.MockArgoCDAppClientCalledWith{}
			reconciler.ArgoCDAppClient = &mockedArgoCDAppClient
```
I think I'm missing why we changed this

maruina (commented):
Looks good, the code duplication is well worth fixing the race condition.

controllers/progressivesync_controller_test.go
sledigabel: good catch, it's prob a mess with my rebase.

maruina (approved):
LGTM

Issue #63, opened 2021-04-19
Add `progressDeadlineSeconds`


maruina:
To make sure the operator doesn't get stuck in an infinite loop, we need a safeguard mechanism. We want this to be eventual consistent, because there might be a delay in ArgoCD (it might be down, or unable to sync for a time window).

We can use the `progressDeadlineSeconds` concept as seen in https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#deploymentspec-v1-apps and https://docs.flagger.app/usage/how-it-works

maruina:
This is a good example about how this is used: https://github.com/fluxcd/flagger/blob/main/pkg/canary/daemonset_ready.go#L74

maruina:
Considering how ArgoCD support retries with backoff, as per https://argo-cd.readthedocs.io/en/stable/operator-manual/application.yaml
```
    # The retry feature is available since v1.7
    retry:
      limit: 5 # number of failed sync attempt retries; unlimited number of attempts if less than 0
      backoff:
        duration: 5s # the amount to back off. Default unit is seconds, but could also be a duration (e.g. "2m", "1h")
        factor: 2 # a factor to multiply the base duration after each failed retry
        maxDuration: 3m # the maximum amount of time allowed for the backoff strategy

```
we can tune ArgoCD to retry how many times we need with the appropriate backoff.

This behavior makes the `ProgressDeadline` feature less urgent.

PR #87, opened 2021-05-24
Reuse context in reconciler

### LINKED ISSUE #67: Pass context in reconciliation methods instead of defining a new one
We have few methods of our `ProgressiveRolloutReconciler` structure which they need a context.

Instead of using the context defined inside the `Reconcile` method, we re-define it as a background context. This mean we lose all the information about the deadlines if we ever need to cancel the context for whatever reason.

We should instead propagate the context to the dependent methods

DiogoMCampos:
Fixes #67.

Two questions:
- I haven't touched `requestsForApplicationChange` nor `requestsForSecretChange` as they are called in `handler.EnqueueRequestsFromMapFunc(r.requestsForApplicationChange)` and `handler.EnqueueRequestsFromMapFunc(r.requestsForSecretChange)` respectively. Should we?
- I'm reusing the global `ctx` in tests - is there any better approach?

### DIFF
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -101,7 +101,7 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		log = r.Log.WithValues("stage", stage.Name)
 
 		// Get the clusters to update
-		clusters, err := r.getClustersFromSelector(stage.Targets.Clusters.Selector)
+		clusters, err := r.getClustersFromSelector(ctx, stage.Targets.Clusters.Selector)
 		if err != nil {
 			message := "unable to fetch clusters"
 			log.Error(err, message)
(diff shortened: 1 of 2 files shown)


sledigabel (approved):
LGTM, but do we know where is the `ctx` for the `Reconcile` function coming from and who sets it?
the logic being, it might be a context that has nothing to do with the actual context we want propagated (in anticipation of #63)

sledigabel:
> Fixes #67.
> 
> Two questions:
> 
> * I haven't touched `requestsForApplicationChange` nor `requestsForSecretChange` as they are called in `handler.EnqueueRequestsFromMapFunc(r.requestsForApplicationChange)` and `handler.EnqueueRequestsFromMapFunc(r.requestsForSecretChange)` respectively. Should we?
> * I'm reusing the global `ctx` in tests - is there any better approach?


OK So THAT is the answer to my question, the context is initialised there :-)
I think it's ok we can leave that like that for now.

@maruina could prob confirm this.

maruina:
I think we should change `requestsForApplicationChange` and `requestsForSecretChange` as they define they own context (see for example https://github.com/Skyscanner/applicationset-progressive-sync/pull/87/files#diff-a853c97824440aa3801077e1f5661f9087e27091372140f15f89ed5c8d0c5ba6R247)

Regarding the test, I'm not sure about that so I think you can keep the current `ctx`.

sledigabel:
> I think we should change `requestsForApplicationChange` and `requestsForSecretChange` as they define they own context

But if they don't generate the context, what will?
It seems to be that this is the context that gets propagated to the Reconciler?
Presumably the context needs to be instantiated somewhere?

maruina:
> > I think we should change `requestsForApplicationChange` and `requestsForSecretChange` as they define they own context
> 
> But if they don't generate the context, what will?
> It seems to be that this is the context that gets propagated to the Reconciler?
> Presumably the context needs to be instantiated somewhere?

You are correct. I was already thinking in a kubebuilder 3.0 scenario.

maruina (approved):
LGTM

aravindvnair99 (approved):

PR #90, opened 2021-05-27
Release a new chart version

nebojsa-prodana:
The crd definition has been renamed in #84 so we need to package a new chart that can install the correct crd definitions.


DimitarHristov111 (commented):
Bumps the chart version.

DimitarHristov111 (approved):

PR #91, opened 2021-06-01
Fix stage status phase field serialization to json

nebojsa-prodana:


dmcampos (approved):
Fixes serialized name of `Phase` field. LGTM!

PR #92, opened 2021-06-01
Fix release workflow

nebojsa-prodana:
See: https://github.com/olegtarasov/get-tag/issues/14


sledigabel (approved):
LGTM, shouldn't that get covered by dependabot?

nebojsa-prodana:
> LGTM, shouldn't that get covered by dependabot?

Yeah I was thinking the same. For some reason it is missing this one...

PR #96, opened 2021-06-08
Fix status update

### LINKED ISSUE #93: Reset ProgressiveSync object
The ProgressiveSync object needs to be reset when there's a change in stages status. This applies to both `stages` status - which could be reset/cleaned - and to `conditions` - which should be set to reflect the new overall status of the object.

An example is a completed progressive sync (with condition `Complete` set to `True`) that starts synchronizing again and sets a stage to be `Progressing`. The latest event's `Complete` condition should not be `True` at this point.

nebojsa-prodana:
Fixes #93

### DIFF
--- a/api/v1alpha1/condition_types.go
+++ b/api/v1alpha1/condition_types.go
@@ -25,9 +25,13 @@ const (
 	// the progressive sync stages have been completed
 	StagesCompleteReason string = "StagesCompleted"
 
-	// StagesCompleteReason represent the fact that all
-	// the progressive sync stages have been completed
+	// StagesInitializedReason represent the fact that all
+	// the progressive sync stages have been initialized
 	StagesInitializedReason string = "StagesInitialized"
+
+	// StagesProgressingReason represent the fact that some
+	// of the progressive sync stages are progressing
+	StagesProgressingReason string = "StagesProgressing"
 
 	// StagesFailedReason represent the fact that a
 	// progressive sync stage has failed
--- a/api/v1alpha1/condition_types.go
+++ b/api/v1alpha1/condition_types.go
@@ -25,10 +25,6 @@ const (
 	// the progressive sync stages have been completed
 	StagesCompleteReason string = "StagesCompleted"
 
-	// StagesInitializedReason represent the fact that all
-	// the progressive sync stages have been initialized
-	StagesInitializedReason string = "StagesInitialized"
-
 	// StagesProgressingReason represent the fact that some
 	// of the progressive sync stages are progressing
 	StagesProgressingReason string = "StagesProgressing"
--- a/api/v1alpha1/condition_types.go
+++ b/api/v1alpha1/condition_types.go
@@ -25,6 +25,10 @@ const (
 	// the progressive sync stages have been completed
 	StagesCompleteReason string = "StagesCompleted"
 
+	// StagesCompleteReason represent the fact that all
+	// the progressive sync stages have been completed
+	StagesInitializedReason string = "StagesInitialized"
+
 	// StagesFailedReason represent the fact that a
 	// progressive sync stage has failed
 	StagesFailedReason string = "StagesFailed"
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -71,7 +71,7 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	if err := r.initializeStatus(ctx, &pr); err != nil {
-		log.Error(err, "unable to clear status")
+		log.Error(err, "unable to initialize status")
 		return ctrl.Result{}, err
 	}
 
@@ -111,9 +111,13 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if err != nil {
 			message := "unable to fetch clusters"
 			log.Error(err, message)
-			if err := r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr); err != nil {
+			r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr)
+
+			if err := r.submitStatus(ctx, &pr); err != nil {
+				r.Log.Error(err, "failed to update object status")
 				return ctrl.Result{}, err
 			}
+
 			return ctrl.Result{}, err
 		}
 		r.Log.V(1).Info("clusters selected", "clusters", utils.GetClustersName(clusters.Items))
@@ -123,9 +127,13 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if err != nil {
 			message := "unable to fetch apps"
 			log.Error(err, message)
-			if err := r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr); err != nil {
+			r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr)
+
+			if err := r.submitStatus(ctx, &pr); err != nil {
+				r.Log.Error(err, "failed to update object status")
 				return ctrl.Result{}, err
 			}
+
 			return ctrl.Result{}, err
 		}
 		r.Log.V(1).Info("apps selected", "apps", utils.GetAppsName(apps))
@@ -136,7 +144,11 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if err := r.removeAnnotationFromApps(ctx, outOfSyncApps, utils.ProgressiveSyncSyncedAtStageKey); err != nil {
 			message := "unable to remove out-of-sync annotation"
 			log.Error(err, message)
-			if err := r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr); err != nil {
+
+			r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr)
+
+			if err := r.submitStatus(ctx, &pr); err != nil {
+				r.Log.Error(err, "failed to update object status")
 				return ctrl.Result{}, err
 			}
 			return ctrl.Result{}, err
@@ -155,9 +167,16 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 			if err != nil {
 				if !strings.Contains(err.Error(), "another operation is already in progress") {
-					message := "unable to sync app"
+					message := "unable to remove out-of-sync annotation"
 					log.Error(err, message, "message", err.Error())
-					if err := r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr); err != nil {
+
+					r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr)
+
+					// TODO: Should this be reflected in the general condition?
+					// failed := pr.NewStatusCondition(syncv1alpha1.CompletedCondition, metav1.ConditionTrue, syncv1alpha1.StagesFailedReason, message)
+					// apimeta.SetStatusCondition(pr.GetStatusConditions(), failed)
+					if err := r.submitStatus(ctx, &pr); err != nil {
+						r.Log.Error(err, "failed to update object status")
 						return ctrl.Result{}, err
 					}
 
@@ -169,16 +188,14 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 
 		if scheduler.IsStageFailed(apps) {
-			message := "stage failed"
-			log.Info(message)
-			if err := r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseFailed, &pr); err != nil {
-				return ctrl.Result{}, err
-			}
-
 			failedMessage := fmt.Sprintf("%s stage failed", stage.Name)
+			log.Info(failedMessage)
+
+			r.updateStageStatus(ctx, stage.Name, failedMessage, syncv1alpha1.PhaseFailed, &pr)
+
 			failed := pr.NewStatusCondition(syncv1alpha1.CompletedCondition, metav1.ConditionTrue, syncv1alpha1.StagesFailedReason, failedMessage)
 			apimeta.SetStatusCondition(pr.GetStatusConditions(), failed)
-			if err := r.setStatus(ctx, &pr); err != nil {
+			if err := r.submitStatus(ctx, &pr); err != nil {
 				r.Log.Error(err, "failed to update object status")
 				return ctrl.Result{}, err
 			}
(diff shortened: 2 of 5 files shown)


controllers/progressivesync_controller.go
```diff
@@ -224,7 +230,7 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 // SetupWithManager adds the reconciler to the manager, so that it gets started when the manager is started.
 func (r *ProgressiveSyncReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&syncv1alpha1.ProgressiveSync{}).
+		For(&syncv1alpha1.ProgressiveSync{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
```
nebojsa-prodana, line 233: See: https://github.com/operator-framework/operator-sdk/issues/2795

This effectively makes the reconciler not reconcile on its own status changes.

api/v1alpha1/condition_types.go
```diff
@@ -25,6 +25,14 @@ const (
 	// the progressive sync stages have been completed
 	StagesCompleteReason string = "StagesCompleted"
 
+	// StagesInitializedReason represent the fact that all
+	// the progressive sync stages have been initialized
+	StagesInitializedReason string = "StagesInitialized"
```
maruina, line 30: What is the difference between `StagesInitialized` and `StagesProgressing`? How do we distinguish between them?

controllers/progressivesync_controller.go
```diff
@@ -418,12 +431,44 @@ func (r *ProgressiveSyncReconciler) syncApp(ctx context.Context, appName string)
...
+		return nil
+
+	})
+	return retryErr
+}
+
+// initializeStatus initializes the progressive sync object status to prevent conflicts when updating status later on
+func (r *ProgressiveSyncReconciler) initializeStatus(ctx context.Context, pr *syncv1alpha1.ProgressiveSync) error {
+
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
maruina, line 458: Why do we need the retry policy here? I thought that getting the latest object first was enough to avoid conflicts.

controllers/progressivesync_controller.go
```diff
@@ -418,12 +431,44 @@ func (r *ProgressiveSyncReconciler) syncApp(ctx context.Context, appName string)
...
+
+		key := client.ObjectKeyFromObject(pr)
+		latest := syncv1alpha1.ProgressiveSync{}
+		if err := r.Client.Get(ctx, key, &latest); err != nil {
+			return err
+		}
+		latest.Status = pr.Status
+		if err := r.Client.Status().Update(ctx, &latest); err != nil {
```
maruina, line 445: Why are we dropping the `r.Client.Status().Patch(ctx, pr, client.MergeFrom(&latest))` and use `Status().Update` instead?

internal/scheduler/scheduler.go
```diff
@@ -61,7 +63,11 @@ func Scheduler(apps []argov1alpha1.Application, stage syncv1alpha1.ProgressiveSy
 		return scheduledApps
 	}
 
-	for i := 0; i < maxParallel-len(progressingApps); i++ {
+	// It may be that at point when apps were observed, no apps were progressing yet
+	// so maxParallel-len(progressingApps) might actually be greater than len(outOfSyncApps)
+	// causing the runtime to panic
```
maruina, line 68: Do you have an example of this behaviour? Is there any test we can add so we can be sure we're solving that bug?

maruina (commented):
Some questions but looks promising.

internal/scheduler/scheduler.go
nebojsa-prodana: It might be hard to write test for it, it happened to me whilst using a debugger because the debugger can pause all the execution while the rest of the world (ArgoCD) moves on. Basically what happened was:

```
maxParallel-len(progressingApps)
2
len(outOfSyncApps)
1
```

So there were no progressingApps at the moment, probably because they already got synced while the debugger was paused, but there was still a stale outOfSyncApps that was retrieved earlier on. I've repeated it several times in debugger, progressively stopping the world for less and less time, so it seems like this is a possible race condition that can happen.

controllers/progressivesync_controller.go
nebojsa-prodana: We don't need to merge anymore, because we already have the latest state retrieved, so we can just update instead.

controllers/progressivesync_controller.go
nebojsa-prodana: We probably don't need it anymore, initially I had this method setting the state on the server side as well, we no longer do that since the reconcilation loop will eventually in all it's code paths prior to requeueing push the state to the server.

I'll see if I can remove this, nice catch! 👍

api/v1alpha1/condition_types.go
nebojsa-prodana: Indeed, I wasn't sure whether we'd need something like this. StagesInitialized is just the empty/default state with no per stage status, while StagesProgressing was to be set as soon as the controller makes progress on one of the stages. It might be useful in detecting whether the controller is progressing at all, but like I said, I am not sure myself if it is needed right now. Happy to remove it and keep things a bit simpler in the code as well.

controllers/progressivesync_controller.go
```diff
@@ -143,7 +149,7 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		// TODO: status status - update targets
 
 		for _, s := range scheduledApps {
-			r.Log.Info("syncing app", "app", s)
+			log.Info("syncing app", "app", s)
```
sledigabel, line 152: Why is this changing?

nebojsa-prodana:
```
log := r.Log.WithValues("ProgressiveSync", req.NamespacedName)
```

I figured it would be more useful to log to which instance of progressivesync object do these logs pertain to.

api/v1alpha1/condition_types.go
maruina: I would start with just Progressing and then we can expand it in the future.

controllers/progressivesync_controller.go
```diff
@@ -418,12 +431,44 @@ func (r *ProgressiveSyncReconciler) syncApp(ctx context.Context, appName string)
...
+		}
+		pr = &latest
+		return nil
+
+	})
+	return retryErr
+}
+
+// initializeStatus initializes the progressive sync object status to prevent conflicts when updating status later on
+func (r *ProgressiveSyncReconciler) initializeStatus(ctx context.Context, pr *syncv1alpha1.ProgressiveSync) error {
```
sledigabel, line 456: semantics, but perhaps `initializeStatus` should be `resetStatus`? and the outcome of `reset` would be `initialized`?

internal/scheduler/scheduler.go
```diff
@@ -61,7 +63,11 @@ func Scheduler(apps []argov1alpha1.Application, stage syncv1alpha1.ProgressiveSy
...
+	p := (int)(math.Min((float64)(maxParallel-len(progressingApps)), (float64)(len(outOfSyncApps))))
```
sledigabel, line 69: I'd suggest clarifying this line with a if rather than the double casting. It's relatively simple to be explicited.
```
p := maxParallel-len(progressingApps)
if p > len(outofSyncApps){
  p = len(outOfSyncApps)
}
```

controllers/progressivesync_controller.go
sledigabel: my bad, i missed this.

controllers/progressivesync_controller.go
nebojsa-prodana: This was removed in a0f7606772c9e9a6482961ed62463a238f4a3f71, it was not needed!

nebojsa-prodana:
> * Useless assignment to local variable

Seems like a false alert to me, the pr object is updated through pass by reference in the caller's code.

controllers/progressivesync_controller.go
maruina: While this unblock us for the time being, I'd like to raise a separate issue to have a long term fix for this because now the controller can't self-heal if something changes the status.

controllers/progressivesync_controller.go
```diff
@@ -418,12 +460,40 @@ func (r *ProgressiveSyncReconciler) syncApp(ctx context.Context, appName string)
 	return r.ArgoCDAppClient.Sync(ctx, &syncReq)
 }
 
-// patchStatus patches the progressive sync object status
-func (r *ProgressiveSyncReconciler) patchStatus(ctx context.Context, pr *syncv1alpha1.ProgressiveSync) error {
+// submitStatus updates the progressive sync object status with backoff
+func (r *ProgressiveSyncReconciler) submitStatus(ctx context.Context, pr *syncv1alpha1.ProgressiveSync) error {
```
maruina, line 464: Why not `updateStatus`?

controllers/progressivesync_controller.go
```diff
@@ -418,12 +460,40 @@ func (r *ProgressiveSyncReconciler) syncApp(ctx context.Context, appName string)
...
+
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
maruina, line 466: Are we still need the retry here?

controllers/progressivesync_controller.go
```diff
@@ -105,9 +111,13 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if err != nil {
 			message := "unable to fetch clusters"
 			log.Error(err, message)
-			if err := r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr); err != nil {
+			r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseUnknown, &pr)
+
+			if err := r.submitStatus(ctx, &pr); err != nil {
+				r.Log.Error(err, "failed to update object status")
```
maruina, line 117:
```suggestion
				log.Error(err, "failed to update object status")
```

sledigabel:
Quick summary of our conversation:
1) refactor the min calculation
2) check out all the `r.Log` and see if they can just be `log.`
3) refactor the `initializeStatus` to be closer to the format of the one in the helm controller: https://github.com/fluxcd/helm-controller/blob/e25d6894bfdc547101e60dbaf739f99ab4f005ef/controllers/helmrelease_controller.go#L184
4) same for `submitStatus`

3 and 4 should fix the problem we're seeing at the linting phase.
@maruina could pick this up in @nebojsa-prodana 's absence

controllers/progressivesync_controller.go
```diff
@@ -418,12 +312,135 @@ func (r *ProgressiveSyncReconciler) syncApp(ctx context.Context, appName string)
...
+			return err
+		}
+		return nil
+
+	})
+	return retryErr
+}
+
+// reconcileStage reconcile a ProgressiveSyncStage
+func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv1alpha1.ProgressiveSync, stage syncv1alpha1.ProgressiveSyncStage) (syncv1alpha1.ProgressiveSync, reconcile.Result, error, bool) {
```
DiogoMCampos, line 336: I'm not a fan of returning four different values, to be honest. I'm aware that it's idiomatic Go to return two values for error-handling, but having three other return values makes this difficult to follow. 

I think as a starting point, named return values (if at all possible?) and definitely adding documentation would make this bettre, as it's not immediately clear what the `bool` value means. 

Further refactoring to avoid 3 non-error return values would probably be beneficial, but I understand if we'd rather do that later. :slightly_smiling_face:

controllers/progressivesync_controller.go
DiogoMCampos: Slightly tangential: do we know why these conflicts happen?

controllers/progressivesync_controller.go
DiogoMCampos: A +1 to raising this as a separate issue.

controllers/progressivesync_controller.go
```diff
@@ -394,19 +293,14 @@ func (r *ProgressiveSyncReconciler) removeAnnotationFromApps(ctx context.Context
 }
 
 // updateStageStatus updates the target stage given a stage name, message and phase
-func (r *ProgressiveSyncReconciler) updateStageStatus(ctx context.Context, name, message string, phase syncv1alpha1.StageStatusPhase, pr *syncv1alpha1.ProgressiveSync) error {
+func (r *ProgressiveSyncReconciler) updateStageStatus(ctx context.Context, name, message string, phase syncv1alpha1.StageStatusPhase, pr *syncv1alpha1.ProgressiveSync) {
```
DiogoMCampos, line 296: Why have we removed the `error` return value here?

controllers/progressivesync_controller.go
DiogoMCampos: Well, I just saw https://github.com/fluxcd/helm-controller/blob/e25d6894bfdc547101e60dbaf739f99ab4f005ef/controllers/helmrelease_controller.go#L184 that Seb linked and can see they have 3 returns values (2 + error) - so this isn't a 'first' :slightly_smiling_face: 

I'm still not overly fond of multiple return values, but at least naming and/or documenting them something would be a good first step. 

Truth is, the whole reconciliation loop has multiple responsibilities - or at least a wide scope of responsibility -, making it harder to have a single return value.

controllers/progressivesync_controller.go
maruina: Done: https://github.com/Skyscanner/applicationset-progressive-sync/issues/99

controllers/progressivesync_controller.go
maruina: In the end we decided to leave the retry logic and do a full status update. This is the recommended approach suggested in https://github.com/kubernetes/client-go/blob/master/examples/create-update-delete-deployment/main.go#L125

controllers/progressivesync_controller.go
maruina: I was able to remove a variable because we can use `result.Requeue` to see if we need to requeue or not.

I don't think we can remove the ProgressiveSync variable unless we start doing things with pointer, but we decided to follow a value semantic.

Are you happy with the latest implementation? Is there anything else do you think we can improve?

controllers/progressivesync_controller.go
maruina: I think it's because `pr.SetStageStatus` doesn't return anything and we now make the update while calling `updateStatusWithRetry`.

api/v1alpha1/condition_types.go
sledigabel, line 30: We don't appear to need this any longer?

controllers/progressivesync_controller.go
DiogoMCampos: Yea, looks good :+1:

controllers/progressivesync_controller.go
```diff
@@ -60,171 +61,69 @@ type ProgressiveSyncReconciler struct {
...
-			apimeta.SetStatusCondition(pr.GetStatusConditions(), failed)
-			if err := r.patchStatus(ctx, &pr); err != nil {
-				r.Log.Error(err, "failed to update object status")
-				return ctrl.Result{}, err
-			}
-			r.Log.Info("rollout failed")
-			// We can set Requeue: true once we have a timeout in place
-			return ctrl.Result{}, nil
+		if result.Requeue {
+			log.Info("requeuing stage")
```
sledigabel, line 106: Nitpicking here, but wonder whether this is relevant as `Info` from a logging perspective?
This feels like `Debug` to me.

controllers/progressivesync_controller.go
```diff
@@ -60,171 +61,69 @@ type ProgressiveSyncReconciler struct {
 
 // Reconcile performs the reconciling for a single named ProgressiveSync object
 func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	log := r.Log.WithValues("ProgressiveSync", req.NamespacedName)
+	log := r.Log.WithValues("progressivesync", req.NamespacedName)
```
sledigabel, line 64: Why did we change the case here?

maruina: To be consistent with https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller.go#L72 where we say `applicationset`. I thought having logs in lower-case would it be better.

api/v1alpha1/condition_types.go
maruina: Good catch, thanks

maruina: Removed

controllers/progressivesync_controller.go
maruina: Looking at the information we currently show at the `Info` level, it feels it gives you the full picture. We already say when an app is sync and what's the stage status.
I'm happy to move it to a different level if we thing it's more a debug info.

sledigabel: we can still adjust it if the info logs become too verbose.

sledigabel (approved):
LGTM
well done all, great opportunity to fix a bug and refactor.

PR #97, opened 2021-06-08
Add more scripts to ease local debug, dev and testing

nebojsa-prodana:
Supersedes https://github.com/Skyscanner/applicationset-progressive-sync/pull/58

This adds:
* Additional params to the argo secrets to allow us to write ProgressiveSync objects with more than one stage for local testing
* Scripts for deploying a test appset and a test prog rollout
   * We are missing a reliable way of moving the needle to test the actual updates. So far I have been manually editing the applications, but we should instead make several different releases with different version values of https://github.com/maruina/appset-poc/blob/main/goinfra/Chart.yaml#L6 to allow is to easily flip-flop between them
* Make targets for spinning up the debugger
* vscode configuration for attaching to debugger
  * Configuration is missing for goland. I am not adept at using it, so I didn't have time to test it in there


hack/debug.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+root=$(dirname "${BASH_SOURCE[0]}")
```
DimitarHristov111, line 4: I just tried this locally is it always returns `.`. Is this expected and it it is always `.` do we need it declared this way?

hack/debug.sh
```diff
@@ -0,0 +1,10 @@
...
+
+# shellcheck disable=SC1091
+source .env.local
+
+bash "$root/redeploy-dev-resources.sh"
+dlv --listen=:2345 --headless=true --api-version=2 debug main.go -- --zap-devel=true
```
DimitarHristov111, line 10: Is the `--` after main.go necessary?

hack/deploy-test-appset.sh
```diff
@@ -1,13 +1,16 @@
...
-set -ex
+set -e
 
 root=$(dirname "${BASH_SOURCE[0]}")
 
 prevcontext=$(kubectl config current-context)
 kubectl config use-context kind-argocd-control-plane
 
-# TODO: Make this generate argo apps in all created clusters
+kubectl get applications -n argocd --no-headers | cut -d' ' -f1 | xargs kubectl -n argocd delete application
```
DimitarHristov111, line 9: Should we have `|| echo "Not found"` similarly to how we delete the application set below?

hack/dev-functions.sh
```diff
@@ -43,6 +43,18 @@ function register_argocd_cluster() {
 	kubectl config use-context "$prevcontext"
 }
 
+function label_argocd_cluster() {
```
DimitarHristov111, line 46: Sorry if it is a silly question, but how does the labels fit with the debug configuration? I looked at the other changes, but couldn't see the connection between them 😞

DimitarHristov111 (commented):
Overall looks good, left some clarifying questions.

hack/debug.sh
nebojsa-prodana: yes, it passes args to the actual go process being forked by dlv

hack/deploy-test-appset.sh
nebojsa-prodana: We are getting application names, so unless something else deletes them in the meantime, I think this is fine as is.

hack/dev-functions.sh
nebojsa-prodana: The labeling affects how the progressive sync controller is going to apply https://github.com/Skyscanner/applicationset-progressive-sync/pull/97/files#diff-adb910ba444459702e453de3f6ed05206831fe068d92bc236e44e1bfe6303a81R1

So when as part of `make debug` we deploy the test prog sync object, we will sort of end up having a stable repeatable env for local testing/debugging.

hack/debug.sh
nebojsa-prodana: This makes it sort of independent of the `pwd` from which the script was invoked.

DimitarHristov111: I see, ok, my idea was that if the value is always the current directory `./`, why don't we just use the literal `./`.

README.md
```diff
@@ -174,19 +174,44 @@ After running the script, you will have 3 kind clusters created locally:
...
+
+``` shell
+make debug
+```
+
+from another terminal session to kill the debugger.
+
+### Debugging tests
+
+Delve can be used to debug tests as well. See `Test` launch configuration in `.vscode/launch.json`. Something similar should be achievable in your IDE of choice as well.
```
DimitarHristov111, line 214: As the rest of us use GoLand, could we add some instructions for it and mainly what we did yesterday. Essentially:

Update the make test step in the make file to be:
```
# Run tests
test: tools generate fmt vet manifests
	dlv test ./controllers
```

and set breakpoint inside the code using `runtime.Breakpoint()` and use the `dvl help` to see how the debugger CLI works.

hack/deploy-test-appset.sh
sledigabel, line 9: Have you tried `--all` in the `kubectl delete` ?

hack/deploy-test-progsync.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -e
+
+root=$(dirname "${BASH_SOURCE[0]}")
+
+prevcontext=$(kubectl config current-context)
+kubectl config use-context kind-argocd-control-plane
+
+# Remove finalizer to allow deletion to complete
+(kubectl get progressivesync -n argocd goinfra -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -) || echo "Continuing.."
```
sledigabel, line 10: have you tried to use `kubectl patch` here?

sledigabel (commented):
a few comments

README.md
nebojsa-prodana: Those instructions are not Goland specific, i.e they would work with any editor. I can add them, but they are more of hacks than instructions. I'd rather spend sometime down the line for adding proper integration for Goland.

hack/deploy-test-appset.sh
nebojsa-prodana: Ah, good call, I kept it like this while testing the script as it allowed me to try and delete a single application at a time instead of nuking everything. I don't see why `kubectl delete --all` wouldn't work here. 👍

hack/deploy-test-progsync.sh
nebojsa-prodana: Done!

.vscode/launch.json
```diff
@@ -0,0 +1,42 @@
...
+        },
+        {
+            "name": "Unit tests - scheduler",
+            "type": "go",
+            "request": "launch",
+            "mode": "test",
+            "program": "${workspaceFolder}/internal/scheduler",
+        }
+        {
+            "name": "Unit tests - utils",
```
nebojsa-prodana, line 35: Go extension is a bit annoying, it doesn't allow you to specify recursive debugging, so we either have to set it like this or change the target program as needed in case that we want to debug some unit tests.

README.md
DimitarHristov111: Ok, I agree.

DimitarHristov111 (approved):
Adds and improves our scripts to allow for creating of a dev environment. This way, we will be able to more easily test and debug it locally.

PR #100, opened 2021-06-14
Fix release process

maruina:
Looking at the `Release Artifacts` it hasn't been running for a while. This PR:
- replace a third-party action with the docker one to extract metadata for tags. You can see it in action here: https://github.com/users/maruina/packages/container/package/docker-withings-sync (see `Recent tagged image versions`)
- use the git tag for the docker tag
- improve the release drafter template to leverage our issue labels


dmcampos (approved):
LGTM! Fixes automated release 👍

PR #101, opened 2021-06-14
Update release-artifacts.yaml

DiogoMCampos:


sledigabel:
I'm going to need a bit of background on this

DiogoMCampos:
Sorry - created this on a whim after we noticed the current approach wasn't working. I think this may be a bug with GitHub (or Actions, perhaps), to be honest. **tl;dr:** publishing a release that was previously a draft from `github-actions` does not trigger a `push tags` event.

See https://github.com/release-drafter/release-drafter/issues/876 for a more detailed description.

sledigabel (approved):
LGTM, will change the event type to a published release, so we can use the UI rather than pushing a tag

DimitarHristov111 (approved):
Updates the GitHub action config to match the suggested solution here - https://github.com/release-drafter/release-drafter/issues/876. This should fix the issue of not running the action when a new release is created from a draft.

PR #102, opened 2021-06-14
Should fix the race condition once and for all

sledigabel:
... on #81.
It's fixed by keeping a local copy on the test scope, leaving no chance
for `namespace` to change in the process.

Begone, race condition of hell! 😈

### DIFF
--- a/controllers/progressivesync_controller_test.go
+++ b/controllers/progressivesync_controller_test.go
@@ -81,13 +81,13 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 
 		It("should forward events for owned applications", func() {
 			By("creating an owned application")
-
-			ownerPR := createOwnerPR(namespace, "owner-pr")
+			localNS := namespace
+			ownerPR := createOwnerPR(localNS, "owner-pr")
 			Expect(k8sClient.Create(ctx, ownerPR)).To(Succeed())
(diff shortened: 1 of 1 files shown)


sledigabel:
For posterity:
image

controllers/progressivesync_controller_test.go
```diff
@@ -81,13 +81,13 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 
 		It("should forward events for owned applications", func() {
 			By("creating an owned application")
-
-			ownerPR := createOwnerPR(namespace, "owner-pr")
+			localNS := namespace
```
sledigabel, line 84: the alternative to the local copy is to create the `types.NamespacedName` comparison variable _before_ the `Eventually`

DiogoMCampos (commented):
Why should we use a local namespace specifically in here, but not other tests?

sledigabel:
> Why should we use a local namespace specifically in here, but not other tests?

It is the only test that reuses the namespace in the `expect`
https://github.com/Skyscanner/applicationset-progressive-sync/pull/102/files#diff-d75f956a084c139d4d522260c50a218c56e711fbc18c42f930907e27c890c654R107

DiogoMCampos (approved):
Uses local namespace variable so we avoid global variable changes failing a test on occasion. LGTM!

Issue #103, opened 2021-06-15
Infinite loop when multiple stage targeting the same clusters


maruina:
We got an infinite loop with the following ProgressiveSync

```
spec:
[...]
  stages:
  - maxParallel: 1
    maxTargets: 1
    name: one cluster as canary in eu-west-1
    targets:
      clusters:
        selector:
          matchLabels:
            region: eu-west-1
  - maxParallel: 3
    maxTargets: 3
    name: one cluster as canary in every other region
    targets:
      clusters:
        selector:
          matchExpression:
          - key: region
            operator: NotIn
            values:
            - eu-west-1
    name: rollout to remaining clusters
  - maxParallel: 25%
    maxTargets: 100%
    targets:
      clusters:
        selector: {}
```

After some investigation, we found the following issues:
- The scheduler relies on an annotation to determine when an Application has been synced (see https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/internal/scheduler/scheduler.go#L24). While we check for the annotation, we never set it. Because of that, we never hit this code path: https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/internal/scheduler/scheduler.go#L59
- We always re-write the stage status, but looking at https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/api/v1alpha1/progressivesync_types.go#L107:28 we always overwrite the `StartedAt` and `FinishedAt` time
- We didn't catch this scenario in our test because our test at https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller_test.go#L268 has a label selector so we return only one cluster per stage. We should change the second stage to have a label selector to include the first cluster as well.

PR #104, opened 2021-06-17
Add a more real case scenario reconcile test

maruina:
- As discussed in #103 , the https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller_test.go#L268 test has a very naive ProgressiveSync object: two stages with a selector that returns only one cluster, and targeting a single cluster at the time.

We should change this test to better reflect a real case scenario. We've internally used as default the ProgressiveSync reported in #103 so this PR change the test to use that ProgressiveSync.

In addition, I changed a couple of helpers function to be more generic for when we want to add more tests with different combination of clusters.

**TODO**

- [x] ~Fix MaxTargets to return a percentange of the remaining Out Of Sync apps instead of all the apps~ we will address this in https://github.com/Skyscanner/applicationset-progressive-sync/issues/112 because it's a quite big change
- [x] Decomment scheduler tests
- [ ] Decomment removeAnnotation

### DIFF
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -19,7 +19,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"github.com/argoproj/gitops-engine/pkg/health"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


DiogoMCampos:
There's still some work to do on the multi-stage test, I will take a look at this again tomorrow. We've integrated the failing stage test, but haven't improved it much compared to how it already was.

controllers/progressivesync_controller.go
```diff
@@ -388,32 +396,34 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
...
-
-		r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseFailed, &ps)
-		// Set ProgressiveSync status
-		failed := ps.NewStatusCondition(syncv1alpha1.CompletedCondition, metav1.ConditionFalse, syncv1alpha1.StagesFailedReason, message)
-		apimeta.SetStatusCondition(ps.GetStatusConditions(), failed)
-
-		return ps, ctrl.Result{}, err
-	}
+	//outOfSyncApps := utils.GetAppsBySyncStatusCode(apps, argov1alpha1.SyncStatusCodeOutOfSync)
+	//healthyApps := utils.GetAppsByHealthStatusCode(outOfSyncApps, health.HealthStatusHealthy)
```
maruina, line 407: This condition needs to be `not Progressing` as we want to give a bad deployment time to recovery

internal/scheduler/scheduler.go
```diff
@@ -5,11 +5,12 @@ import (
 	"github.com/Skyscanner/applicationset-progressive-sync/internal/utils"
 	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/gitops-engine/pkg/health"
+	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // Scheduler returns a list of apps to sync for a given stage
-func Scheduler(apps []argov1alpha1.Application, stage syncv1alpha1.ProgressiveSyncStage) []argov1alpha1.Application {
+func Scheduler(log logr.Logger, apps []argov1alpha1.Application, stage syncv1alpha1.ProgressiveSyncStage) []argov1alpha1.Application {
```
maruina, line 13: Maybe we should pass a context here and create a new logger from context. This mean we need to add the keys we care about to the context (how?)

internal/utils/utils.go
```diff
@@ -39,6 +39,22 @@ func GetAppsBySyncStatusCode(apps []argov1alpha1.Application, code argov1alpha1.
 	return result
 }
 
+// GetAppsBySyncAtAnnotation returns the Applications having the specified annotation
+func GetAppsBySyncAtAnnotation(apps []argov1alpha1.Application, annotation string, stageName string) []argov1alpha1.Application {
```
maruina, line 43:
```suggestion
func GetAppsByAnnotation(apps []argov1alpha1.Application, annotation string, stageName string) []argov1alpha1.Application {
```

maruina:
Close this one as we continued the work in #114

PR #105, opened 2021-06-17
Add syncedAt annotation

DimitarHristov111:
Work addresses part of https://github.com/Skyscanner/applicationset-progressive-sync/issues/103

### DIFF
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -332,7 +332,8 @@ func (r *ProgressiveSyncReconciler) updateStatusWithRetry(ctx context.Context, p
 	return retryErr
 }
 
-func (r *ProgressiveSyncReconciler) addSyncedAtAnnotation(ctx context.Context, app argov1alpha1.Application, stageName string) error {
+// setSyncedAtAnnotation sets the SyncedAt annotation for the currently progressing stage of the app
+func (r *ProgressiveSyncReconciler) setSyncedAtAnnotation(ctx context.Context, app argov1alpha1.Application, stageName string) error {
 	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		key := client.ObjectKeyFromObject(&app)
 		latest := argov1alpha1.Application{}
@@ -340,7 +341,10 @@ func (r *ProgressiveSyncReconciler) addSyncedAtAnnotation(ctx context.Context, a
 			return err
 		}
 		val, ok := app.Annotations[utils.ProgressiveSyncSyncedAtStageKey]
+		// Required due to the use of `omitempty` in serialisation.
+		// `omitempty` converts the empty map into nil.
 		if !ok {
+			r.Log.Info("Initialising empty map for Annotations")
 			latest.Annotations = make(map[string]string)
 		}
 		if val != stageName {
@@ -423,9 +427,9 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
 			}
 		}
 
-		err = r.addSyncedAtAnnotation(ctx, s, stage.Name)
+		err = r.setSyncedAtAnnotation(ctx, s, stage.Name)
 		if err != nil {
-			message := "failed to syncedAt annotation"
+			message := "failed to add syncedAt annotation"
 			log.Error(err, message, "message", err.Error())
 
 			return ps, ctrl.Result{}, err
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -281,8 +281,8 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
 // removeAnnotationFromApps remove an annotation from the given Applications
 func (r *ProgressiveSyncReconciler) removeAnnotationFromApps(ctx context.Context, apps []argov1alpha1.Application, annotation string) error {
 	for _, app := range apps {
-		if _, ok := app.ObjectMeta.Annotations[annotation]; ok {
-			delete(app.ObjectMeta.Annotations, annotation)
+		if _, ok := app.Annotations[annotation]; ok {
+			delete(app.Annotations, annotation)
 			if err := r.Client.Update(ctx, &app); err != nil {
 				r.Log.Error(err, "failed to update Application", "app", app.Name)
 				return err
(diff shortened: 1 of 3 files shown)


controllers/progressivesync_controller_test.go
```diff
@@ -88,6 +88,10 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "app",
 					Namespace: localNS,
+					Annotations: map[string]string{
```
maruina, line 91: Out of curiosity, why are we adding those annotations? As default you can have none.

DimitarHristov111: I think initialising the empty map didn't work and was erroring with
>assignment to entry in nil map

However, maybe I didn't initialise it correctly 🤔

DimitarHristov111: This should now be fixed and the initial map should be empty.

controllers/progressivesync_controller.go
```diff
@@ -332,6 +332,28 @@ func (r *ProgressiveSyncReconciler) updateStatusWithRetry(ctx context.Context, p
 	return retryErr
 }
 
+func (r *ProgressiveSyncReconciler) addSyncedAtAnnotation(ctx context.Context, app argov1alpha1.Application, stageName string) error {
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
DiogoMCampos, line 336: This ends up being very similar to https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller.go#L318. Do we think there is an opportunity for some refactoring here?

controllers/progressivesync_controller.go
```diff
@@ -332,6 +332,28 @@ func (r *ProgressiveSyncReconciler) updateStatusWithRetry(ctx context.Context, p
...
+		key := client.ObjectKeyFromObject(&app)
+		latest := argov1alpha1.Application{}
+		if err := r.Client.Get(ctx, key, &latest); err != nil {
+			return err
+		}
+		val, ok := app.Annotations[utils.ProgressiveSyncSyncedAtStageKey]
+		if !ok {
```
DiogoMCampos, line 343: I am generally not a big fan of comments, but this feels like a place where we should add context on why this is here and explain how the `omitempty` works. If I were a first-time code reader, I'd certainly question this `if` clause here.

controllers/progressivesync_controller.go
```diff
@@ -332,6 +332,28 @@ func (r *ProgressiveSyncReconciler) updateStatusWithRetry(ctx context.Context, p
...
 }
 
+func (r *ProgressiveSyncReconciler) addSyncedAtAnnotation(ctx context.Context, app argov1alpha1.Application, stageName string) error {
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		key := client.ObjectKeyFromObject(&app)
+		latest := argov1alpha1.Application{}
+		if err := r.Client.Get(ctx, key, &latest); err != nil {
+			return err
+		}
+		val, ok := app.Annotations[utils.ProgressiveSyncSyncedAtStageKey]
```
DiogoMCampos, line 342: Actually, on  a second look, didn't you push the changes to make this `app.ObjectMeta.Annotations`?

controllers/progressivesync_controller.go
```diff
@@ -400,6 +422,14 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
 				return ps, ctrl.Result{}, err
 			}
 		}
+
+		err = r.addSyncedAtAnnotation(ctx, s, stage.Name)
+		if err != nil {
+			message := "failed to syncedAt annotation"
```
DiogoMCampos, line 428:
```suggestion
			message := "failed to add syncedAt annotation"
```

DiogoMCampos (commented):
Left a couple of comments, but otherwise all looks good to me.

controllers/progressivesync_controller.go
nebojsa-prodana: This is manipulating argocd apps though, I think it is fine like this, i.e it is fine to build on top of RetryOnConflict.

I would argue though that this is not just adding the annotation it is also updating it if the stage name has changed, so maybe it should be called `setSyncedAtAnnotation` instead? Also, we are missing the comment above the method.

controllers/progressivesync_controller.go
```diff
@@ -332,6 +332,28 @@ func (r *ProgressiveSyncReconciler) updateStatusWithRetry(ctx context.Context, p
...
+		key := client.ObjectKeyFromObject(&app)
+		latest := argov1alpha1.Application{}
+		if err := r.Client.Get(ctx, key, &latest); err != nil {
+			return err
+		}
+		val, ok := app.Annotations[utils.ProgressiveSyncSyncedAtStageKey]
+		if !ok {
+			latest.Annotations = make(map[string]string)
+		}
+		if val != stageName {
```
nebojsa-prodana, line 346: Is it worth logging something that this has happened for any potential debugging purposes later on?

controllers/progressivesync_controller.go
nebojsa-prodana: Alternatively, maybe this can be made a part of syncApp method instead.

DimitarHristov111: I am more in favour of keeping it as a separate function as it will allow for easier unit tests. 

The question for refactoring - there are similarities but also they change and update different elements. The only part that I can see being repeated is when we retrieve the latest app object. So I can take this part out in a separate function, what do you think?

For the function name, yes, I agree, I will update it.

controllers/progressivesync_controller.go
DimitarHristov111: No, I reverted them as it seems that it cane be done this way as well. I decided to leave it consistent with the rest of the code.

DiogoMCampos: Ah - alright, TIL!

controllers/progressivesync_controller.go
DiogoMCampos: Good points - agree with renaming it but keeping the separate function.

DiogoMCampos (approved):
Adds syncedAt annotation. The tests don't change much, but we'll use the test from #104. LGTM!

controllers/progressivesync_controller.go
sledigabel: Sounds like this thread should be resolved :-)

DimitarHristov111: Oh, yes, sorry, resolving now. 🙂

sledigabel:
LGTM, quick question though: what happens if an app is in two stages?
Does it get retagged with `syncAt`? will this affect anything?

DimitarHristov111:
>LGTM, quick question though: what happens if an app is in two stages?
Does it get retagged with syncAt? will this affect anything?

I spoke with Nebojsa about it, and we shouldn't get the app in two stages anymore. And more specifically, the check should happen around here - https://github.com/Skyscanner/applicationset-progressive-sync/blob/adcdbcad865ae90f6ab5fc0adcd13390c61bc6fe/internal/scheduler/scheduler.go#L37

maruina:
An app can be selected in multiple stages, because the label selector could
catch it, but should be synced only in one stage. After that, as Dimitar
pointed out, it should be ignored by the scheduler.

sledigabel:
Brilliant. thanks

sledigabel (approved):
LGTM

PR #107, opened 2021-06-23
Make release-artifacts push artifacts

nebojsa-prodana:


DimitarHristov111 (approved):
Adds the `push` input value as described in the docs here - https://github.com/docker/build-push-action#inputs. This will allows us to push the image to dockerhub.

PR #108, opened 2021-06-23
Trigger image release on release being published

nebojsa-prodana:
We got releases working off manual tags in #107, now let us try to restore the functionality of #101 which should allow us to get a new tagged image once release drafter has published the release.

### DIFF
--- a/.github/workflows/release-artifacts.yaml
+++ b/.github/workflows/release-artifacts.yaml
@@ -1,9 +1,8 @@
 name: Release Artifacts
 
 on:
-  push:
-    tags:
-      - 'v*'
+  release:
+    types: [published]
 
 jobs:
   release:
@@ -34,5 +33,5 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: true
-          tags: ${{ steps.meta.outputs.tags }}
+          tags: ${{ github.event.release.tag_name }}
           labels: ${{ steps.meta.outputs.labels }}


.github/workflows/release-artifacts.yaml
```diff
@@ -34,5 +33,5 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: true
-          tags: ${{ steps.meta.outputs.tags }}
+          tags: ${{ github.event.release.tag_name }}
```
nebojsa-prodana, line 36: See: https://stackoverflow.com/questions/59518658/how-to-use-github-release-version-number-in-github-action/59521156#59521156

DimitarHristov111: This way, we will not have the `latest` and `commit_sha` tags. Could this break something, somewhere? Just making sure that we have checked and we are sure 🙂

sledigabel:
LGTM.

Re/ the trigger event, we use the same thing over in turbolift:
https://github.com/Skyscanner/turbolift/blob/main/.github/workflows/release.yml

sledigabel (approved):

.github/workflows/release-artifacts.yaml
nebojsa-prodana: We don't have the latest anyways, so I think this is fine.

PR #109, opened 2021-06-24
Add update and patch to the permissions

DimitarHristov111:
Ref: Adding `update` and `patch` permissions to allow us to add the syncedAt annotation to the application object.


dmcampos (approved):
Adds permissions to update and patch argoproj.io applications. Needed for #105 to work. LGTM!

PR #110, opened 2021-06-24
Use chart version that doesn't exist

DimitarHristov111:
Use chart version which is not present here: https://api.github.com/repos/Skyscanner/applicationset-progressive-sync/releases.

Background: At the moment, we seem to release a new chart using the latest tag on the repo, which is not good. I will add a task to fix this. For now, we can just use a chart number which doesn't exist yet.


nebojsa-prodana (approved):
LGTM!

We should eventually sort out the release process. The chart release action is looking for the latest tag to make the release which might be out of sync with the chart version inside of Chart.yml. To avoid this from happening in the future we should release both the image and the chart upon baking the release.

Issue #112, opened 2021-06-30
Support maxTargets and maxParallel with % values


maruina:
The ProgressiveSync CRD is designed as to support `maxTargets` and `maxParallel` as [IntOrString](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/intstr#IntOrString).

Consider the following scenario, when you have 4 clusters and a ProgressiveSync similar to
```
[...]
stages:
  - name: one cluster
     maxTargets: 1
     maxParallel: 1
     targets:
       clusters:
         selector:
           name: cluster-one
  - name: everything else
     maxTargets: 100%
     maxParallel: 25%
     targets:
       clusters:
         selector: {}
```

In the second stage, we would expect maxTargets to be `3`, but instead we set it to `4`. This is because [the scheduler looks at every application matching the label selector](https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller.go#L411) - in this case all of them.

We need to change this logic so when we express maxTargets and maxParallel as %, they are scaled against the _remaining_ clusters.

Issue #111, opened 2021-06-28
Scope the controller to a single namespace


maruina:
We coded the controller under the assumption that is running in a single namespace but we've seen instances where we have multiple Applications or ProgressiveSync under different namespaces interfering with each other (#103).
We also want to run multiple controller versions in the same cluster under different namespaces so we can test a new version without impacting every customer.

However, we didn't do the proper setup as per https://book.kubebuilder.io/cronjob-tutorial/empty-main.html#every-journey-needs-a-start-every-program-a-main. In particular, we're not passing the required `Namespace` to the manager and we might also want to watch a specific set of namespaces only.

We need two CLI flags to specify two different namespaces:
- the namespace where the controller is running (default to the ArgCD default namespace, `argocd`)
- the namespace where ArgoCD is running, since we need to watch its Applications and Secrets

When we do this change we also need to change the Helm Chart because if it's namespace scoped we don't need a ClusterRole and a ClusterRoleBinding but just a Role and a RoleBinding.

maruina:
We've seen also this issue when deploying the controller in out internal environment. We have multiple argocd targeting multplie clusters, and using the label selector was picking up duplicated clusters.

PR #114, opened 2021-07-02
Fix flaky CI with a complex test case

### LINKED ISSUE #103: Infinite loop when multiple stage targeting the same clusters

nebojsa-prodana:
Fix #103 

- As discussed in #103 , the https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller_test.go#L268 test has a very naive ProgressiveSync object: two stages with a selector that returns only one cluster, and targeting a single cluster at the time.

We should change this test to better reflect a real case scenario. We've internally used as default the ProgressiveSync reported in #103 so this PR change the test to use that ProgressiveSync.

In addition, I changed a couple of helpers function to be more generic for when we want to add more tests with different combination of clusters.

**TODO**

- [x] ~Fix MaxTargets to return a percentange of the remaining Out Of Sync apps instead of all the apps~ we will address this in https://github.com/Skyscanner/applicationset-progressive-sync/issues/112 because it's a quite big change
- [x] Decomment scheduler tests
- [x] Decomment removeAnnotation

This is a fork of #104 on commit cab8fff024c43a064ae7aab359f14caa14ff0f7a

### DIFF
--- a/.github/workflows/test-build.yaml
+++ b/.github/workflows/test-build.yaml
@@ -38,16 +38,29 @@ jobs:
       - name: Build
         run: make manager
 
-      - name: Login to GitHub Packages Docker Registry
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            ghcr.io/skyscanner/applicationset-progressive-sync
+          # generate Docker tags based on the following events/attributes
+          tags: |
+            type=ref,event=pr
+            type=semver,pattern={{raw}}
+            type=sha
+
+      - name: Login to GitHub Packages
         uses: docker/login-action@v1.9.0
         with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
-        if: github.event_name != 'pull_request'
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build and publish a docker image when merge in main
+      - name: Build and publish a docker image
         uses: docker/build-push-action@v2.5.0
         with:
           push: true
-          tags: maruina/applicationset-progressive-sync:main
-        if: github.event_name != 'pull_request'
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ lint: fmt vet
 
 # Run tests
 test: tools generate fmt vet manifests
-	ginkgo -r --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
+	ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
 
 # Build manager binary
 manager: generate fmt vet
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ lint: fmt vet
 
 # Run tests
 test: tools generate fmt vet manifests
-	ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
+	ginkgo -r --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
 
 # Build manager binary
 manager: generate fmt vet
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ lint: fmt vet
 
 # Run tests
 test: tools generate fmt vet manifests
-	ginkgo -r --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
+	ginkgo -reportPassed -r --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
 
 # Build manager binary
 manager: generate fmt vet
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ lint: fmt vet
 
 # Run tests
 test: tools generate fmt vet manifests
-	ginkgo -reportPassed -r --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
+	ginkgo -r --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
 
 # Build manager binary
 manager: generate fmt vet
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -62,20 +62,21 @@ type ProgressiveSyncReconciler struct {
 // Reconcile performs the reconciling for a single named ProgressiveSync object
 func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithValues("progressivesync", req.NamespacedName)
+	log.Info("Starting reconciliation loop")
 
 	// Get the ProgressiveSync object
-	var pr syncv1alpha1.ProgressiveSync
-	if err := r.Get(ctx, req.NamespacedName, &pr); err != nil {
+	var ps syncv1alpha1.ProgressiveSync
+	if err := r.Get(ctx, req.NamespacedName, &ps); err != nil {
 		log.Error(err, "unable to fetch progressivesync object")
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	log = r.Log.WithValues("applicationset", pr.Spec.SourceRef.Name)
+	log = log.WithValues("applicationset", ps.Spec.SourceRef.Name)
 
 	// If the object is being deleted, remove finalizer and don't requeue it
-	if !pr.ObjectMeta.DeletionTimestamp.IsZero() {
-		controllerutil.RemoveFinalizer(&pr, syncv1alpha1.ProgressiveSyncFinalizer)
-		if err := r.Update(ctx, &pr); err != nil {
+	if !ps.ObjectMeta.DeletionTimestamp.IsZero() {
+		controllerutil.RemoveFinalizer(&ps, syncv1alpha1.ProgressiveSyncFinalizer)
+		if err := r.Update(ctx, &ps); err != nil {
 			log.Error(err, "failed to update object when removing finalizer")
 			return ctrl.Result{}, err
 		}
@@ -84,9 +85,9 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 	}
 
 	// Add finalizer if it doesn't exist
-	if !controllerutil.ContainsFinalizer(&pr, syncv1alpha1.ProgressiveSyncFinalizer) {
-		controllerutil.AddFinalizer(&pr, syncv1alpha1.ProgressiveSyncFinalizer)
-		if err := r.Update(ctx, &pr); err != nil {
+	if !controllerutil.ContainsFinalizer(&ps, syncv1alpha1.ProgressiveSyncFinalizer) {
+		controllerutil.AddFinalizer(&ps, syncv1alpha1.ProgressiveSyncFinalizer)
+		if err := r.Update(ctx, &ps); err != nil {
 			log.Error(err, "failed to update object when adding finalizer")
 			return ctrl.Result{}, err
 		}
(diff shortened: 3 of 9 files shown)


maruina:
We found multiple business logic issues:
- https://github.com/Skyscanner/applicationset-progressive-sync/pull/114/commits/40824ef7ca25a167a0dfb963f0c73eebeee1947e when we now re-schedule for sync the progressing application as well so we can annotate them. This protect us against a case when someone is manually syncing from the ArgoCD UI and we were not annotating the app
- https://github.com/Skyscanner/applicationset-progressive-sync/pull/114/commits/40824ef7ca25a167a0dfb963f0c73eebeee1947e we were missing a corner case in the scheduling logic 
- if the labels selector returns an empty selection, meaning `apps == nil`, we decided to make the stage as completed

Makefile
```diff
@@ -23,7 +23,7 @@ lint: fmt vet
 
 # Run tests
 test: tools generate fmt vet manifests
-	ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
+	ginkgo -reportPassed -r --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
```
maruina, line 26: We need to remove `-reportPassed before merging

internal/scheduler/scheduler.go
```diff
@@ -5,11 +5,12 @@ import (
 	"github.com/Skyscanner/applicationset-progressive-sync/internal/utils"
 	argov1alpha1 "github.com/argoproj/argo-cd/pkg/apis/application/v1alpha1"
 	"github.com/argoproj/gitops-engine/pkg/health"
+	"github.com/go-logr/logr"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 // Scheduler returns a list of apps to sync for a given stage
-func Scheduler(apps []argov1alpha1.Application, stage syncv1alpha1.ProgressiveSyncStage) []argov1alpha1.Application {
+func Scheduler(log logr.Logger, apps []argov1alpha1.Application, stage syncv1alpha1.ProgressiveSyncStage) []argov1alpha1.Application {
```
maruina, line 13: We need to remove the logging before merging, or keep it as it is depending on what people feels about this.

Raise #115 to properly address this.

DimitarHristov111 (commented):
We reviewed it together yesterday and the logic seems fine. We still need to verify that it works as expected using the snapshot image. There are a few comments that came up during the review and I left them below.

controllers/progressivesync_controller_test.go
```diff
@@ -231,230 +242,501 @@ var _ = Describe("ProgressiveRollout Controller", func() {
...
+										Values:   []string{"eu-west-1"},
+									}},
+								},
+							}},
+					}, {
+						Name: "rollout to remaining clusters",
+						// MaxParallel: intstr.Parse("25%"),
+						// MaxTargets:  intstr.Parse("100%"),
+						MaxParallel: intstr.Parse("1"),
+						MaxTargets:  intstr.Parse("4"),
```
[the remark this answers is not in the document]
maruina: Still doesn't work. We have https://github.com/Skyscanner/applicationset-progressive-sync/issues/112 to fix this

controllers/progressivesync_controller_test.go
```diff
@@ -231,230 +242,501 @@ var _ = Describe("ProgressiveRollout Controller", func() {
...
+			}))
+
+			// Make sure the ProgressiveSync is completed
+			expected := ps.NewStatusCondition(syncv1alpha1.CompletedCondition, metav1.ConditionTrue, syncv1alpha1.StagesCompleteReason, "All stages completed")
+			ExpectCondition(&ps, expected.Type).Should(HaveStatus(expected.Status, expected.Reason, expected.Message))
 
-		It("should fail if unable to sync application", func() {
-			testPrefix := "failed-stages"
+			Expect(k8sClient.Delete(ctx, &ps)).To(Succeed())
+		})
```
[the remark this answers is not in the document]
maruina: Removed

controllers/progressivesync_controller_test.go
```diff
@@ -231,230 +242,501 @@ var _ = Describe("ProgressiveRollout Controller", func() {
...
+			ExpectCondition(&ps, progress.Type).Should(HaveStatus(progress.Status, progress.Reason, progress.Message))
+
+			By("completing 25% of the third stage applications")
+			Eventually(func() error {
+				return setAppStatusCompleted(ctx, "account1-eu-west-1a-2", namespace)
+			}).Should(Succeed())
+
+			// Make sure the current stage is still in progress
+			// because we still have 75% of clusters to sync
+			ExpectStageStatus(ctx, psKey, "rollout to remaining clusters").Should(MatchStage(syncv1alpha1.StageStatus{
```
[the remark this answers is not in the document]
maruina: Done

.github/workflows/test-build.yaml
```diff
@@ -38,16 +38,29 @@ jobs:
...
         run: make manager
 
-      - name: Login to GitHub Packages Docker Registry
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          # list of Docker images to use as base name for tags
+          images: |
+            ghcr.io/skyscanner/applicationset-progressive-sync
```
maruina, line 47: We're now pushing to Github packages when opening a PR. This allow us to pull a specific docker image for e2e testing before merging it into main.

Makefile
```diff
@@ -23,7 +23,7 @@ lint: fmt vet
 
 # Run tests
 test: tools generate fmt vet manifests
-	ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
+	ginkgo -r --failOnPending --cover -coverprofile=../coverage.out --trace --race --progress
```
maruina, line 26: Remove the randomize options at least until we address https://github.com/Skyscanner/applicationset-progressive-sync/issues/111
Until the controller isn't scoped to a single namespace, the ginkgo random options might cause different objects to pollute the test namespace.

controllers/progressivesync_controller.go
```diff
@@ -388,32 +383,17 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
...
-		apimeta.SetStatusCondition(ps.GetStatusConditions(), failed)
-
-		return ps, ctrl.Result{}, err
+		return ps, ctrl.Result{Requeue: true, RequeueAfter: requeueDelayOnError}, err
 	}
+	log.Info("fetched apps targeting selected clusters", "apps", utils.GetAppsName(apps))
 
 	// Get the Applications to update
-	scheduledApps := scheduler.Scheduler(apps, stage)
+	scheduledApps := scheduler.Scheduler(log, apps, stage)
```
maruina, line 391: The scheduler logging as proved to be quite valuable. I propose we keep it in this PR and we fix-forward in https://github.com/Skyscanner/applicationset-progressive-sync/issues/115

controllers/progressivesync_controller_test.go
```diff
@@ -745,10 +1104,10 @@ func TestSync(t *testing.T) {
 
 	testAppName := "foo-bar"
 
-	application, error := r.syncApp(ctx, testAppName)
+	application, err := r.syncApp(ctx, testAppName)
```
maruina, line 1107: The IDE was complaining that `error` is a built-in package

controllers/progressivesync_controller.go
```diff
@@ -211,14 +212,15 @@ func (r *ProgressiveSyncReconciler) requestsForSecretChange(o client.Object) []r
...
 				/*
-					Consider the following scenario:
-					- 2 Applications
-					- owned by the same ApplicationSet
-					- referenced by the same Progressive Rollout
-					- targeting the same cluster
-
-					In this scenario, we would trigger the reconciliation loop twice.
-					To avoid that, we use a map to store for which Progressive Rollout objects we already trigger the reconciliation loop.
+						Consider the following scenario:
```
sledigabel, line 215: did `go fmt` triggered that tabbing here?

controllers/progressivesync_controller.go
```diff
@@ -334,34 +322,41 @@ func (r *ProgressiveSyncReconciler) updateStatusWithRetry(ctx context.Context, p
...
-			r.Log.Info("Initialising empty map for Annotations")
-			latest.Annotations = make(map[string]string)
-		}
-		if val != stageName {
-			r.Log.Info("Setting the annotation for the currently progressing stage")
-			latest.Annotations[utils.ProgressiveSyncSyncedAtStageKey] = stageName
-			if err := r.Client.Update(ctx, &latest); err != nil {
-				return err
+
+		log = log.WithValues("app", fmt.Sprintf("%s/%s", app.Namespace, app.Name))
```
sledigabel, line 337: could we use latest instead of apps here?

controllers/progressivesync_controller.go
maruina: Yes, I'm not sure why to be honest.

maruina: I can revert it if we want.

controllers/progressivesync_controller.go
maruina: Fixed

controllers/progressivesync_controller.go
sledigabel: i was just curious :-)

controllers/progressivesync_controller.go
```diff
@@ -388,32 +383,17 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
...
 
 	// Get the Applications to update
-	scheduledApps := scheduler.Scheduler(apps, stage)
+	scheduledApps := scheduler.Scheduler(log, apps, stage)
 
 	for _, s := range scheduledApps {
-		log.Info("syncing app", "app", s)
+		log.Info("syncing app", "app", fmt.Sprintf("%s/%s", s.Namespace, s.Name), "sync.status", s.Status.Sync.Status, "health.status", s.Status.Health.Status)
 
-		_, err := r.syncApp(ctx, s.Name)
```
sledigabel, line 416: why was that one removed and another one at line 412 added?
If it's the same it is safer to leave the walrus operator and make sure the errors don't persist within blocks

internal/utils/utils.go
```diff
@@ -58,15 +74,15 @@ func GetSyncedAppsByStage(apps []argov1alpha1.Application, name string) []argov1
 
 	for _, app := range apps {
 		val, ok := app.Annotations[ProgressiveSyncSyncedAtStageKey]
-		if ok && val == name && app.Status.Sync.Status == argov1alpha1.SyncStatusCodeSynced && app.Status.Health.Status == health.HealthStatusHealthy {
+		if ok && val == name && app.Status.Sync.Status == argov1alpha1.SyncStatusCodeSynced {
```
sledigabel, line 77: what was the rationale behind this removal?

sledigabel (commented):
Some small comments, overall LGTM

internal/utils/utils.go
maruina: Because the function is called `GetSyncedAppsByStage` so it should get `synced` app in the given stage - meaning it has the annotation.
The old function was also looking for the health status and we didn't need that.

controllers/progressivesync_controller.go
```diff
@@ -334,34 +322,41 @@ func (r *ProgressiveSyncReconciler) updateStatusWithRetry(ctx context.Context, p
...
 		return nil
 	})
 	return retryErr
 }
 
 // reconcileStage reconcile a ProgressiveSyncStage
 func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv1alpha1.ProgressiveSync, stage syncv1alpha1.ProgressiveSyncStage) (syncv1alpha1.ProgressiveSync, reconcile.Result, error) {
-	log := r.Log.WithValues("stage", stage.Name)
+	log := r.Log.WithValues("progressivesync", fmt.Sprintf("%s/%s", ps.Namespace, ps.Name), "applicationset", ps.Spec.SourceRef.Name, "stage", stage.Name)
+	requeueDelayOnError := time.Minute * 5
```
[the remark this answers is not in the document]
maruina: True, made it a constant

controllers/progressivesync_controller.go
maruina: Good catch, I restored the previous commit

sledigabel (approved):
LGTM!

PR #116, opened 2021-07-06
Store the annotated apps in memory

DimitarHristov111:
Builds on top of https://github.com/Skyscanner/applicationset-progressive-sync/pull/114. The aim is to help us see if storing the apps annotation in memory will fix the infinite loop.


controllers/progressivesync_controller.go
```diff
@@ -349,6 +357,23 @@ func (r *ProgressiveSyncReconciler) setSyncedAtAnnotation(ctx context.Context, a
...
+		appKeyValue := ps.Name + "/" + app.Name
+
+		r.SyncedAtStage[appKeyValue] = stageName
+
+		val, ok := r.SyncedAppsPerStage[stageKeyValue]
+		if !ok {
+			r.SyncedAppsPerStage[stageKeyValue] = 1
+		} else {
+			if val+1 > stageMaxTargets {
+				r.SyncedAppsPerStage[stageKeyValue] = stageMaxTargets
```
DiogoMCampos, line 371: I'm a bit afraid this could mean we  'hide' that we actually synced more apps than we should - but I have no proof to back this claim, so perhaps something we can accept and keep an eye on?

DiogoMCampos (changes requested):
Beyond comments I left, just a note to remove the Annotations from all Applications test cases.

Overall LGTM - I'm on the fence about some changes, but we can iterate if we find problems.

controllers/progressivesync_controller_test.go
```diff
@@ -822,6 +821,225 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 			expected := failedStagePS.NewStatusCondition(syncv1alpha1.CompletedCondition, metav1.ConditionFalse, syncv1alpha1.StagesFailedReason, "stage 0 stage failed")
 			ExpectCondition(&failedStagePS, expected.Type).Should(HaveStatus(expected.Status, expected.Reason, expected.Message))
 		})
+
+		It("should set conditions when apps already synced", func() {
```
DimitarHristov111, line 825: I wonder if we still need this test as it seems very similar to the one below 🤔

DimitarHristov111 (commented):
It includes quite a lot of changes for a single PR but from what I can see it looks good. It is better than before for sure 🙂  We don't have the tests for what happens when the controller restarts, but I can add it. Based on the current implementation, I'd expect it to still pass as you seem to have already added the implementation to fix it. I will remove it from my side and only add the test.

controllers/progressivesync_controller.go
```diff
@@ -97,11 +99,16 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	latest := ps
```
sledigabel, line 102: why do we need `latest`? looks like you're constantly reusing it after that, I see no more ref to `ps`

sledigabel: line 107

controllers/progressivesync_controller.go
```diff
@@ -388,13 +374,15 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
 	}
 	log.Info("fetched apps targeting selected clusters", "apps", utils.GetAppsName(apps))
 
+	pss.RefreshState(apps, stage)
```
sledigabel, line 377: reconciling the view of the state.
goes through all the apps and figure out what's synced and what's not.

internal/utils/state.go
```diff
@@ -0,0 +1,189 @@
...
+	defer s.Mutex.Unlock()
+
+	unmarkedApps := s.getUnmarkedApps(apps, stage)
+
+	//TODO: While this buys us time for now, as it calculates correctly MaxTargets and MaxParallel
+	// even when they are percentages, this will break if the users have changed those fields in progressive sync spec
+	// because we only initialize it once
+	// We should react to the event of progressive sync spec being changed and invalidate these maps
+	if _, ok := s.MaxParallelPerStage[stage.Name]; !ok {
+		maxTargets, _ := intstr.GetScaledValueFromIntOrPercent(&stage.MaxTargets, len(unmarkedApps), false)
```
sledigabel, line 136: let's talk.

sledigabel (approved):
LGTM.  Creates an interface to store state for synced app and a concrete implementation with an array in memory.

I'd like a follow up conversation about the GetScaledValueFromIntOrPercent with you both and @maruina as discussed.
But this is not covered by this PR.

DiogoMCampos (approved):
Addresses comments above :+1:

PR #119, opened 2021-07-08
Scope the controller to a single namespace

maruina:
Fix https://github.com/Skyscanner/applicationset-progressive-sync/issues/111

I'm not changing the ClusterRole to Role as the kubebuilder documentation suggests when using a single namespace. The reason is because we might have the situation where the controller is deployed in a namespace and argocd in another, so we need to watch two namespaces.


controllers/progressivesync_controller_test.go
```diff
@@ -82,10 +85,34 @@ var _ = Describe("ProgressiveRollout Controller", func() {
...
+			Client:          k8sManager.GetClient(),
+			Log:             ctrl.Log.WithName("controllers").WithName("progressivesync"),
+			ArgoCDAppClient: &mockAcdClient,
+		}
+		err = reconciler.SetupWithManager(k8sManager)
+		Expect(err).ToNot(HaveOccurred())
+
+		go func() {
+			defer GinkgoRecover()
+			err = k8sManager.Start(ctx)
```
maruina, line 109: `k8sManager.Start` takes a context, so we pass the context we created with the cancel option and then send the signal to stop the manager with `defer cancel()` in `AfterEach`

main.go
```diff
@@ -60,8 +63,15 @@ func main() {
 	flag.Parse()
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
+	cacheNamespaces := []string{namespace}
+	if namespace != argocdNamespace {
```
maruina, line 67: This is to avoid creating two informers, two caches, etc. for the same namespace. Happy to change it with a more elegant solution.

sledigabel: Could you give some context around the cache? Seems to be a new addition and not sure what it does.
Also, the argocdNamespace seems to only be used for this check?
Shouldn't we propagate this somewhere? or is it what the cache is doing?

sledigabel: ok, I've read https://pkg.go.dev/sigs.k8s.io/controller-runtime#hdr-Clients_and_Caches and I understand a bit better now. Question: from my understanding we're moving from no cache to using cache. Is this going to cause issues in terms of cache invalidation?

Also, in the context where the controller runs in a different namespace than argocd, will it get notified by the resource changed on the appset or secrets? Sounds to me like the cache is only caching, not giving you the ability to watch another namespace?

maruina: > Question: from my understanding we're moving from no cache to using cache. Is this going to cause issues in terms of cache invalidation?

My understanding from the link above is that when I read
```
The default client provided by the manager reads from a local shared cache (pkg/cache) and writes directly to the API server [...]
```
it means we had the cache even before.
Reading at https://book.kubebuilder.io/cronjob-tutorial/empty-main.html#every-journey-needs-a-start-every-program-a-main and https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder makes me think that instead of caching ProgressiveSync, Applications and Secrets across all the namespace we cache only the objects in those namespaces.

maruina: >  Also, the argocdNamespace seems to only be used for this check?

Basically the idea is "if namespace and argocdNamespace are the same, watch only one namespace, otherwise the two of them"

sledigabel: is adding a namespace as a cacheable space automatically makes it watch it? I would have thought it was under the namespace itself. You're the expert though :-)

maruina:
```
MultiNamespacedCacheBuilder - Builder function to create a new multi-namespaced cache. This will scope the cache to a list of namespaces. Listing for all namespaces will list for all the namespaces that this knows about. By default this will create a global cache for cluster scoped resource. Note that this is not intended to be used for excluding namespaces, this is better done via a Predicate. Also note that you may face performance issues when using this with a high number of namespaces. 
```
I would say yes.
I also added it into the tests and they're passing so it seems to me we can still watch the resources.

maruina: Ah, I see your question now. Besides reading the docs, the only option is to add two namespaces to the tests: the random one + always "argocd". Then we add a test where we create a secret in argoCD and we make sure we see the event.

What do you think?

sledigabel: yeah that works, or we could also decide/enforce the use of the same namespace as argocd and make it unique by design? or is this going to conflict with multiple argocd instances running in the same cluster?

maruina: Unless we drop that requirement so when we update the controller is for everyone :)

sledigabel: that's a strong change of approach but not necessarily a bad idea actually;
in practice the controller would be "bundled" with a running argocd controller so why not making this a strong 1:1 relationship.

How would you feel about that?
And everyone else?

maruina:
commit `eddb362` fix the following error
```
[1625836507] Controller Suite - 9/9 specs STEP: bootstrapping test environment
2021-07-09T15:15:28.490+0200	DEBUG	controller-runtime.test-env	starting control plane	{"api server flags": []}
2021-07-09T15:15:36.444+0200	DEBUG	controller-runtime.test-env	installing CRDs
2021-07-09T15:15:36.447+0200	DEBUG	controller-runtime.test-env	reading CRDs from path	{"path": "../config/crd/bases"}
2021-07-09T15:15:36.501+0200	DEBUG	controller-runtime.test-env	read CRDs from file	{"file": "argoproj.skyscanner.net_progressivesyncs.yaml"}
2021-07-09T15:15:36.538+0200	DEBUG	controller-runtime.test-env	read CRDs from file	{"file": "deployment.skyscanner.net_progressivesyncs.yaml"}
2021-07-09T15:15:36.571+0200	DEBUG	controller-runtime.test-env	read CRDs from file	{"file": "sync.skyscanner.net_progressivesyncs.yaml"}
2021-07-09T15:15:36.571+0200	INFO	controller-runtime.test-env	there are more than one CRD definitions with the same <Group, Version, Kind, Name>	{"GVKN": {"GVK":{"Group":"apiextensions.k8s.io","Version":"v1beta1","Kind":"CustomResourceDefinition"},"Name":"progressivesyncs.argoproj.skyscanner.net"}}
2021-07-09T15:15:36.572+0200	INFO	controller-runtime.test-env	there are more than one CRD definitions with the same <Group, Version, Kind, Name>	{"GVKN": {"GVK":{"Group":"apiextensions.k8s.io","Version":"v1beta1","Kind":"CustomResourceDefinition"},"Name":"progressivesyncs.argoproj.skyscanner.net"}}
2021-07-09T15:15:36.573+0200	DEBUG	controller-runtime.test-env	reading CRDs from path	{"path": "../hack"}
2021-07-09T15:15:36.789+0200	DEBUG	controller-runtime.test-env	read CRDs from file	{"file": "argoproj.io_application.yaml"}
2021-07-09T15:15:36.838+0200	DEBUG	controller-runtime.test-env	installing CRD	{"crd": "progressivesyncs.argoproj.skyscanner.net"}
2021-07-09T15:15:36.870+0200	DEBUG	controller-runtime.test-env	installing CRD	{"crd": "applications.argoproj.io"}
2021-07-09T15:15:36.996+0200	DEBUG	controller-runtime.test-env	adding API in waitlist	{"GV": "argoproj.skyscanner.net/v1alpha1"}
2021-07-09T15:15:36.996+0200	DEBUG	controller-runtime.test-env	adding API in waitlist	{"GV": "argoproj.io/v1alpha1"}
2021-07-09T15:15:37.108+0200	DEBUG	controller-runtime.test-env	installing webhooks
```

maruina:
On second thoughts, it make sense to split this PR so in this one we will scope to a single namespace and we will raise a new one if we want to support multiple namespaces.

main.go
maruina: Happy with that at the moment. The AppSet controller has the same approach

controllers/suite_test.go
```diff
@@ -72,36 +70,15 @@ var _ = BeforeSuite(func(done Done) {
 	Expect(err).ToNot(HaveOccurred())
 	Expect(cfg).ToNot(BeNil())
 
-	err = syncv1alpha1.AddToScheme(scheme.Scheme)
-	err = argov1alpha1.AddToScheme(scheme.Scheme)
-	Expect(err).NotTo(HaveOccurred())
+	Expect(syncv1alpha1.AddToScheme(scheme.Scheme)).To(Succeed())
+	Expect(argov1alpha1.AddToScheme(scheme.Scheme)).To(Succeed())
 
 	// +kubebuilder:scaffold:scheme
```
nebojsa-prodana, line 76: Is kubebuilder scaffolding going to regenerate some of the things that we have removed here?

controllers/suite_test.go
```diff
@@ -72,36 +70,15 @@ var _ = BeforeSuite(func(done Done) {
...
-	})
-	Expect(err).ToNot(HaveOccurred())
-
-	mockAcdClient := mocks.ArgoCDAppClientStub{}
-	reconciler = &ProgressiveSyncReconciler{
-		Client:          k8sManager.GetClient(),
-		Log:             ctrl.Log.WithName("controllers").WithName("progressiverollout"),
-		ArgoCDAppClient: &mockAcdClient,
-	}
-	err = reconciler.SetupWithManager(k8sManager)
```
nebojsa-prodana, line 96: Now that each test in progressive_sync_controller_test.go is going to have its own reconciler, should we add an integration or unit test to prevent regressions on the logic of naming the keys for our in-memory storage #116?

I am saying this because as part of #116, we have added keys for storage that are prefixed by progressive sync name to prevent different tests (different PS in production) from overwriting our storage, with this change to creating a PR once per test instead of per test suite we no longer have a code path that tests that.

controllers/progressivesync_controller_test.go
```diff
@@ -32,7 +35,8 @@ const (
 
 var (
 	appSetAPIRef = utils.AppSetAPIGroup
-	ctx          = context.Background()
+	ctx          context.Context
```
nebojsa-prodana, line 38: I am not the greatest fan of globals, especially if is hard to tell that they are even globals because they have the same naming convention as local variables. I cannot find any guidance for Go about naming the global variables, but I'd personally be happier if we could uppercase them or prefix them with `g` like in C or something.

main.go
```diff
@@ -62,6 +64,8 @@ func main() {
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:             scheme,
+		Namespace:          namespace,
+		NewCache:           cache.MultiNamespacedCacheBuilder([]string{namespace}),
```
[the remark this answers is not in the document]
maruina: I took it from https://book.kubebuilder.io/cronjob-tutorial/empty-main.html#every-journey-needs-a-start-every-program-a-main and reading https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder and https://github.com/argoproj-labs/applicationset/blob/master/main.go#L123

It seems to me that makes sense to restrict the cache since we're watching a single namespace

controllers/suite_test.go
maruina: Not that I'm aware of because I added to it manually

controllers/suite_test.go
maruina: Shouldn't we tackle that in #116 ? This PR doesn't know anything about that structure, no?

main.go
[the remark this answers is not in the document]
maruina: After reading https://github.com/kubernetes-sigs/controller-runtime/blob/v0.9.2/pkg/manager/manager.go#L194 it seems we just need to specify `Namespace`

nebojsa-prodana (approved):
LGTM! Scopes the controller to a single namespace to avoid issues with selecting and watching argocd secrets from different namespaces which can impact the actual rollout schedule.

PR #121, opened 2021-07-20
Add test simulating controller restart

DimitarHristov111:
Builds on top of #116. The aim is to add a test simulating what would happen if the controller restarts.


maruina:
Superseeded by https://github.com/Skyscanner/applicationset-progressive-sync/pull/143

PR #122, opened 2021-07-21
Push release images to GHCR

nebojsa-prodana:


DimitarHristov111 (approved):
Changes the release artifact workflow to publish the image to GitHub Container Registry similar to what we do for the test and build wf here - https://github.com/Skyscanner/applicationset-progressive-sync/blob/7c57da03ba09b363eaf8ea303206d7eecb553cec/.github/workflows/test-build.yaml#L54-L59

LGTM.

PR #123, opened 2021-07-21
Bumps build-push-action version in release-artifacts.yaml

nebojsa-prodana:
Aligns the version to be the same as in https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/.github/workflows/test-build.yaml#L62


DimitarHristov111 (approved):
Makes the version of the build and push action more specific.

PR #124, opened 2021-07-21
Update release-artifacts.yaml

nebojsa-prodana:


DiogoMCampos (approved):
Uses our GHCR repository instead of publishing to Matteo's personal dockerhub one. LGTM

PR #125, opened 2021-07-21
Update release-artifacts.yaml

nebojsa-prodana:


DimitarHristov111 (approved):
Uses the tags from the "meta" step.

PR #130, opened 2021-08-02
Add more logs and fix scheduler logic

DimitarHristov111:
- Adds logs to show more information when the stage is in an unknown state and when an application is changed.
- Fixes the logic inside the scheduler to account for all apps.

### DIFF
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -173,6 +173,8 @@ func (r *ProgressiveSyncReconciler) requestsForApplicationChange(o client.Object
 				Namespace: pr.Namespace,
 				Name:      pr.Name,
 			}})
+			message := fmt.Sprintf("App name: %s; App status: %s", app.Name, &app.Status)
+			r.Log.Info(message)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


controllers/progressivesync_controller.go
```diff
@@ -173,6 +173,8 @@ func (r *ProgressiveSyncReconciler) requestsForApplicationChange(o client.Object
 				Namespace: pr.Namespace,
 				Name:      pr.Name,
 			}})
+			message := fmt.Sprintf("App name: %s; App status: %s", app.Name, &app.Status)
```
maruina, line 176: The whole app status is quite big, maybe we only care about the sync status and health status?

controllers/progressivesync_controller_test.go
```diff
@@ -724,6 +724,36 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 			expected := ps.NewStatusCondition(syncv1alpha1.CompletedCondition, metav1.ConditionTrue, syncv1alpha1.StagesCompleteReason, "All stages completed")
 			ExpectCondition(&ps, expected.Type).Should(HaveStatus(expected.Status, expected.Reason, expected.Message))
 
+			Eventually(func() error {
```
maruina, line 727: Can we add a comment here about why we're extending the test with these assertions?

maruina (commented):
Just a couple of comments but generally LGTM

maruina (dismissed):

maruina (approved):

Issue #131, opened 2021-08-04
Scope the controller to multiple namespaces


maruina:
In https://github.com/Skyscanner/applicationset-progressive-sync/issues/111 we binded the controller to a single namespace. 

There are some cases where we might want to deploy the controller to a different namespace compared to `argocd`. For example. there might be a different team in charge of ArgoCD - and they don't want to give permissions into the `argocd` namespace - or we might want to test multiple version of the controller against the same ArgoCD.

PR #132, opened 2021-08-05
Multi namespace

maruina:
Fix https://github.com/Skyscanner/applicationset-progressive-sync/issues/131

In the tests, I moved the Argo objects inside the `argocd` namespace and left the ProgressiveSync objects inside the random generated namespace. In this way we can test the controller watching multiple namespaces.

### DIFF
--- a/controllers/progressivesync_controller_test.go
+++ b/controllers/progressivesync_controller_test.go
@@ -82,18 +82,18 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 	format.TruncatedDiff = false
 
 	var (
-		namespace string
-		ns        *corev1.Namespace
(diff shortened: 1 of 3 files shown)


controllers/progressivesync_controller_test.go
```diff
@@ -91,8 +92,8 @@ var _ = Describe("ProgressiveRollout Controller", func() {
 		namespace, ns = createRandomNamespace()
```
nebojsa-prodana, line 92: Could we call this `ctrlNamespace` or something to make it a bit clearer?

maruina: Yep, done.

sledigabel (approved):
LGTM

Issue #115, opened 2021-07-05
Improve logging


maruina:
In #114 we needed the Scheduler to log some information. To do that, we passed a `logr.Logger` to the scheduler from the reconciliation loop.

What we need to do:

- [ ] Is that a good idea or do we prefer to add the information we need to the context, pass a context and create a new logger from it? We need to agree on the implementation
- [ ] We only use `log.Info`, while we should agree at least on 2 different logs levels (INFO and DEBUG), their values and start using `log.V().Info` to differentiate between the 2 levels.
- [ ] Add a new flag to enable or disable debug logging or setting a log-level.

maruina:
Fixed in #143

PR #135, opened 2021-09-08
Store the hashed ApplicationSet Spec inside the ProgressiveSync state manager

DimitarHristov111:
## Background
This PR will contribute and help us toward achieving - https://github.com/Skyscanner/applicationset-progressive-sync/issues/63

## Changes
- Fixed the build to correctly download the kubebuilder dep;
- Updated the chart to allow us to `get` and `list` ApplicationSet objects;
- Updated the controller to calculate a hash value from the ApplicationSet spec and store it in memory (for now);
- Added tests;
- Updated the state manager to allow us to store hash values in there;
- Added `github.com/argoproj-labs/applicationset` as a dependency;
- Added `applicationset` as part of the controller scheme;
- Added the ApplicationSet CRD manifest;

## Testing
- Updated tests run as part of the project;
- Tested with a local cluster and did a demo in front of the squad;

## Outstanding tasks
- [ ] Mirror the chart once the PR is merged
- [ ] Deploy the new controller
- [ ] Change the ProgressiveSync scheme to include the namespace as part of the SourceRef section of the manifest.


controllers/progressivesync_controller.go
```diff
@@ -267,6 +280,67 @@ func (r *ProgressiveSyncReconciler) getClustersFromSelector(ctx context.Context,
...
+}
+
+func (r *ProgressiveSyncReconciler) calculateHashValue(ctx context.Context, pr *syncv1alpha1.ProgressiveSync) (string, error) {
+	key := client.ObjectKeyFromObject(pr)
+	latest := syncv1alpha1.ProgressiveSync{}
+	if err := r.Client.Get(ctx, key, &latest); err != nil {
+		return "", err
+	}
+
+	////////////////////////////////////////////////
```
sledigabel, line 319: we prob don't need those comments?

controllers/progressivesync_controller.go
```diff
@@ -267,6 +280,67 @@ func (r *ProgressiveSyncReconciler) getClustersFromSelector(ctx context.Context,
...
+	//return rand.SafeEncodeString(fmt.Sprint(hashValue.Sum32())), nil
+	////////////////////////////////////////////////
+
+	appSet := applicationset.ApplicationSet{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: "argocd", Name: latest.Spec.SourceRef.Name}, &appSet); err != nil {
+		r.Log.Info("Failed to retrieve the ApplicationSet object")
+	}
+
+	hashValue := fnv.New32a()
+	hash.DeepHashObject(hashValue, appSet.Spec)
```
sledigabel, line 339: Don't we need to hash on _both_ the appset and the progressivesync?
If we change the progressivesync parameters it will be considered as a new deployment? (even though, granted, all the apps should be already sync'ed)

controllers/progressivesync_controller.go
DimitarHristov111: Yes, sorry, the PR is still WIP. I think we will want to actually use what's commented as it is more generic and not specific to the ApplicationSet. But I was still working on it, I decided to go with the direct approach of using ApplicationSet for now.

controllers/progressivesync_controller.go
```diff
@@ -99,6 +103,15 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	hashValue, err := r.calculateHashValue(ctx, &ps)
+	if err != nil {
+		log.Info("Failed to generate the hash value from the ApplicationSet spec")
```
sledigabel, line 108: that looks like a blocking error? or do we have a fallback plan?

controllers/progressivesync_controller.go
```diff
@@ -267,6 +289,41 @@ func (r *ProgressiveSyncReconciler) getClustersFromSelector(ctx context.Context,
...
+	}
+
+	apiMetadata := strings.Split(*latest.Spec.SourceRef.APIGroup, "/")
+	genericObj := unstructured.Unstructured{}
+	genericObj.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   apiMetadata[0],
+		Kind:    latest.Spec.SourceRef.Kind,
+		Version: apiMetadata[1],
+	})
+	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: "argocd", Name: latest.Spec.SourceRef.Name}, &genericObj); err != nil {
```
DimitarHristov111, line 308: We need to changed the schema of the ProgressiveSync to also include the namespace inside the SourceRef. This way we will avoid the literal "argocd".

hack/argoproj.io_applicationsets.yaml
```diff
@@ -0,0 +1,6377 @@
+
```
DimitarHristov111, line 1: Taken from https://github.com/argoproj-labs/applicationset/blob/v0.2.0/manifests/crds/argoproj.io_applicationsets.yaml as it is needed for our tests.

controllers/progressivesync_controller.go
DimitarHristov111: This is now removed.

controllers/progressivesync_controller.go
DimitarHristov111: It is just a logging error for now. Once we start working on the next feature, I think it will be more clear what we want to do here.

controllers/progressivesync_controller.go
```diff
@@ -99,6 +103,15 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	hashValue, err := r.calculateHashValue(ctx, &ps)
+	if err != nil {
+		log.Info("Failed to generate the hash value from the ApplicationSet spec")
+	} else {
+		if err := r.updatePsHashAnnotation(ctx, &ps, hashValue); err != nil {
+			log.Info("Failed to update the hash value inside the ProgressiveSync object")
```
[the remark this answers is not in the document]
DimitarHristov111: This is no longer applicable as we no longer update the annotations. I will resolve it and we can discuss the similar point above as it is still relevant there.

controllers/progressivesync_controller.go
maruina: I think this should come from https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/main.go#L55 rather than changing the SourceRef.

controllers/progressivesync_controller.go
maruina: The ProgressiveSync CRD tells you _how_ to sync the applications, so I think we should only hash the ApplicationSet which is the _what_.

controllers/progressivesync_controller.go
```diff
@@ -99,11 +107,25 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{Requeue: true}, nil
 	}
 
+	pss, _ := r.StateManager.Get(ps.Name)
+
+	newHash, err := r.calculateHashedSpec(ctx, &ps)
+	if err != nil {
+		log.Error(err, "Failed to generate the hash value from the ApplicationSet spec")
```
maruina, line 114: Shouldn't we return here?

.github/workflows/test-build.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
           version=2.3.1
           os=$(go env GOOS)
           arch=$(go env GOARCH)
-          curl -Ls https://go.kubebuilder.io/dl/${version}/${os}/${arch} -o /tmp/kubebuilder_${version}.tar.gz
+          curl -Ls https://github.com/kubernetes-sigs/kubebuilder/releases/download/v${version}/kubebuilder_${version}_${os}_${arch}.tar.gz -o /tmp/kubebuilder_${version}.tar.gz
```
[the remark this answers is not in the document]
sledigabel: That would deserve a separate PR I think not to clutter that one.

controllers/progressivesync_controller.go
```diff
@@ -267,6 +289,41 @@ func (r *ProgressiveSyncReconciler) getClustersFromSelector(ctx context.Context,
...
+func (r *ProgressiveSyncReconciler) calculateHashedSpec(ctx context.Context, ps *syncv1alpha1.ProgressiveSync) (string, error) {
+	key := client.ObjectKeyFromObject(ps)
+	latest := syncv1alpha1.ProgressiveSync{}
+	if err := r.Client.Get(ctx, key, &latest); err != nil {
+		r.Log.Error(err, "Failed to retrieve the ProgressiveSync object while calculating the hash value")
+		return "", err
+	}
+
+	apiMetadata := strings.Split(*latest.Spec.SourceRef.APIGroup, "/")
+	genericObj := unstructured.Unstructured{}
```
maruina, line 302: Why do we use the unstructured here? We now that we're getting a reference to the applicationset and we're not supporting a generic `SourceRef`.
Maybe this generate a new discussion about why don't we call it `AppSetRef` instead :)

controllers/progressivesync_controller.go
```diff
@@ -267,6 +289,41 @@ func (r *ProgressiveSyncReconciler) getClustersFromSelector(ctx context.Context,
...
+	objectToHash, ok := genericObj.UnstructuredContent()["spec"]
+	if !ok {
+		err := errors.New("`spec` missing from generic object")
+		r.Log.Error(err, "Missing value from map")
+
+		return "", err
+	}
+	hashValue := fnv.New32a()
+	hash.DeepHashObject(hashValue, objectToHash)
+	r.Log.Info("Successfully calculate the hash value of the Spec")
```
maruina, line 322: Maybe this should be a debug log line instead of info?

controllers/progressivesync_controller.go
```diff
@@ -35,6 +41,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/kubernetes/pkg/util/hash"
```
maruina, line 44: I'm not sure we should have a dependency on `k8s.io` because it's not supported. Why are we trying to do this?

maruina (commented):
A couple of comments but generally looks good!

.github/workflows/test-build.yaml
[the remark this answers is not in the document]
DimitarHristov111: Yeah, we can have it as a separate PR, but I added it here as the whole build for failing to download the dependency.

controllers/progressivesync_controller.go
DimitarHristov111: I was thinking that we can add the logic to return/fail once we start working on the issue for the progress deadline. But I can add it now as well.

controllers/progressivesync_controller.go
DimitarHristov111: Initially we did it by retrieving an ApplicationSet obj, but after discussing it with Nebojsa we agreed that it should be more generic and allow for other object to be used as well. Maybe we should discuss how generic we want it to be.

controllers/progressivesync_controller.go
DimitarHristov111: Hmm, what do you mean that it is not supported? If this is correct (https://pkg.go.dev/k8s.io/kubernetes/pkg/util/hash), there was a new version published recently.

The idea of using `DeepHashObject` was to pass the whole struct object and let it calculate the hash value from it.

.github/workflows/test-build.yaml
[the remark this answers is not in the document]
DimitarHristov111: I have opened a separate PR here - https://github.com/Skyscanner/applicationset-progressive-sync/pull/137 🙂

controllers/progressivesync_controller.go
DimitarHristov111: Not sure I am missing something, but the logger we currently use doesn't have a Debug level of logging - https://pkg.go.dev/github.com/go-logr/logr#pkg-index. There is the idea of levels of verbosity that is set with numbers, but I am not quite sure how this translates to what we want to achieve 🤔

controllers/progressivesync_controller.go
maruina: If you look at https://github.com/kubernetes/kubernetes/issues/81878#issuecomment-696689706 they say that using `k8s.io/kubernetes` is deprecated. Other projects are making an effort to remove that, for example https://github.com/argoproj/gitops-engine/issues/56 and https://github.com/argoproj/argo-cd/issues/5173

At the moment we have a depency on k8s because of argocd, once that is done we don't need that anymore.

controllers/progressivesync_controller.go
maruina: I'm not 100% sure it should be generic since this project is called `applicationset-progressive-sync` :)

.github/workflows/test-build.yaml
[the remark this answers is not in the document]
DimitarHristov111: This is now resolved 🙂

controllers/progressivesync_controller.go
DimitarHristov111: I see, ok, I will try and find a different way for us to calculate the hash value.

DimitarHristov111: If i read [this](https://github.com/argoproj-labs/applicationset/blob/master/api/v1alpha1/applicationset_types.go#L39) correctly, we will get a json from the `spec` [here](https://github.com/Skyscanner/applicationset-progressive-sync/pull/135/files#diff-a853c97824440aa3801077e1f5661f9087e27091372140f15f89ed5c8d0c5ba6R313). And then we can do something like this - https://play.golang.org/p/2clJUvGHJ_l. I will try it tomorrow and see if it works for us.

controllers/progressivesync_controller.go
maruina: You are correct, and we have a ticket to tackle the logging at https://github.com/Skyscanner/applicationset-progressive-sync/issues/115

controllers/progressivesync_controller.go
DimitarHristov111: We had a chat about this and we agreed to look into adding the ProgressiveSync as well, once we have fixed the issue around how we do the calculations with percentages. And also we agreed that it can be done as a separate PR once the fix is in place. Seb, I will resolve for now, but please let us know if there is something you want to discuss.

controllers/progressivesync_controller.go
DimitarHristov111: I have now updated it to return the error we got above. I will resolve for now 🙂

controllers/progressivesync_controller.go
DimitarHristov111: We now use the value of the flag. I have also updated the tests 🙂

controllers/progressivesync_controller.go
DimitarHristov111: I have now updated it to return the error we got above. I will resolve for now 🙂

controllers/progressivesync_controller.go
DimitarHristov111: As we discussed, we agreed to change it back to ApplicationSet.

controllers/progressivesync_controller.go
DimitarHristov111: Ok, thanks, I will resolve as we have an issue to track this.

controllers/progressivesync_controller.go
DimitarHristov111: This dependency is now removed and we rely on internal golang functions now.

controllers/progressivesync_controller.go
```diff
@@ -99,11 +108,26 @@ func (r *ProgressiveSyncReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		return ctrl.Result{}, err
+	} else {
+		currentHash := pss.GetHashedSpec()
+		if currentHash != newHash {
+			pss.SetHashedSpec(newHash)
+			log.Info("Successfully updated the hash value of the ApplicationSet spec", "service", ps.Name)
+		} else {
+			log.Info("Hash value is the same as nothing has changed in the ApplicationSet spec", "service", ps.Name)
+		}
+	}
```
maruina, line 125: I think we should be able to simplify this `if/else`
```suggestion
	if err != nil {
		log.Error(err, "Failed to generate the hash value from the ApplicationSet spec")
		return ctrl.Result{}, err
	}
	currentHash := pss.GetHashedSpec()
	if currentHash != newHash {
		pss.SetHashedSpec(newHash)
		log.Info("Successfully updated the hash value of the ApplicationSet spec", "service", ps.Name)
	} else {
		log.Info("Hash value is the same as nothing has changed in the ApplicationSet spec", "service", ps.Name)
	}
```

controllers/progressivesync_controller.go
```diff
@@ -267,6 +291,35 @@ func (r *ProgressiveSyncReconciler) getClustersFromSelector(ctx context.Context,
...
+
+	appSet := applicationset.ApplicationSet{}
+	if err := r.Client.Get(ctx, client.ObjectKey{Namespace: r.ArgoNamespace, Name: latest.Spec.SourceRef.Name}, &appSet); err != nil {
+		r.Log.Error(err, "Failed to retrieve the ApplicationSet object")
+		return "", err
+	}
+
+	appSetSpec := appSet.Spec
+	appSetSpecInBytes, err := json.Marshal(appSetSpec)
```
maruina, line 310:
```suggestion
	appSetSpecInBytes, err := json.Marshal(appSet.Spec)
```
It seems we only use `appSetSpec` so maybe we can just decode it?

controllers/progressivesync_controller.go
```diff
@@ -267,6 +291,35 @@ func (r *ProgressiveSyncReconciler) getClustersFromSelector(ctx context.Context,
...
+	if err != nil {
+		r.Log.Error(err, "Failed to encode the ApplicanSet spec")
+		return "", err
+	}
+
+	hashedSpecInBytes := md5.Sum(appSetSpecInBytes)
```
maruina, line 316: I don't know if there are any drawbacks about using `md5`. What was the hash function used in kubernetes?

controllers/progressivesync_controller_test.go
```diff
@@ -1127,6 +1254,56 @@ func createApplications(ctx context.Context, targets []Target) ([]argov1alpha1.A
 	return apps, nil
 }
 
+// createApplicationSet creates an ApplicationSet object with labels for the Application template
+func createApplicationSet(ctx context.Context, name string, namespace string, labels map[string]string) (applicationset.ApplicationSet, error) {
```
maruina, line 1258: I'm not sure about this one, but I _think_ it makes it a little more clear. Thoughts?
```suggestion
func createApplicationSetWithLabels(ctx context.Context, name string, namespace string, labels map[string]string) (applicationset.ApplicationSet, error) {
```

controllers/progressivesync_controller.go
DimitarHristov111: I think it boils down to https://github.com/davecgh/go-spew/tree/master/spew, but I struggle to fully understand how they do it. But I don't think it's md5 as there are not references to it. 

About drawbacks, I am not sure. From what I saw, the json is converted to `[]byte` and any changes are visible there. This was enough to convince me that it should be fine and the tests also confirm this. If we want a better understanding, I can spend some more time looking into the `spew` library and try to understand what they do.

DimitarHristov111: I spend a bit more time trying to understand it. It seems that it uses FNV-1 and FNV-1a hash functions (https://pkg.go.dev/hash/fnv#New32a). And then spew is the part that expands the fields and writes into it. Or that's effectively that way I see it - it's a bit confusing. (https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L1165)

If we decide that md5 is not good for our purposes, we can just do what we have here - https://github.com/kubernetes/kubernetes/blob/master/pkg/util/hash/hash.go#L28, but this will introduce a dependency on the spew lib.

dimitar-hristov: I have been thinking about this, and I guess we can do it without spew. I think they use spew so that they can pass in whole objects and extract the data from them (or at least this is my understanding). Spew also does some formatting, which I am not quite sure about. While in our case, at the moment we work with the json directly in the form of `[]byte` given by the `json.Marshal`. However, I am not 100% sure which approach is better.

Unless there is a disadvantage of using the json, we can do something like - https://play.golang.org/p/T3PIfTeAbTt. Line 16 is the one doing the writing, which in kubernetes is done inside spew. But if that's the case, we can use md5 as well, because I can't really see a lot of difference and benefit on one over the other.

This is my understanding at the moment, but I might be missing or misunderstanding something. I see that you are concerned about the use of `md5`, so maybe we can have a more in-depth discussion to clarify what will be best for us.

PS. Sorry for writing from this acc, I hope it's fine :slightly_smiling_face:

maruina: No worries, let's go for md5 for the time being.

maruina (approved):
LGTM

sledigabel (approved):
LGTM, introduces a hash annotation that parses the AppSet and produces a unique hash, and detects a change on the deployment content like that.
The hash is stored in the in-memory store for now.

PR #139, opened 2021-10-15
Migrate to Kubebuilder v3

maruina:
Migrate to kubebuilder v3 following https://book.kubebuilder.io/migration/migration_guide_v2tov3.html

Changes from V2:
- Remove the `tools` directory. In V3 they are using the makefile to ensure controller-gen and kustomize are available with the desired version. We used to do that with the `tools` pattern. We now use the Makefile to download ginkgo as well.
- Moved `SetStageStatus` inside `stage_status_types.go` as it was was the only method related to a stage status in a different file
- Everything under config is generated by kustomize, I just changed the namespace to `argocd` and the prefix to `progressive-sync-`
- Remove the kubebuilder sha checksum as it was causing more issues than benefit


Dockerfile
```diff
@@ -16,13 +16,13 @@ COPY controllers/ controllers/
...
-RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
-USER nonroot:nonroot
+USER 65532:65532
```
sledigabel, line 26: Should we create a user for this? feels weird to use UIDs

Makefile
```diff
@@ -11,65 +11,109 @@ else
...
 
-# Run go vet against code
-vet:
-	go vet ./...
+##@ Build
 
-# Generate code
-generate:
-	controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."
+build: generate fmt vet ## Build manager binary.
```
sledigabel, line 63: I think it would make sense to add `test` in the list?

api/v1alpha1/condition_types.go
```diff
@@ -1,19 +1,3 @@
-/**
```
sledigabel, line 1: Are all the headers gone now?

config/default/kustomization.yaml
```diff
@@ -6,7 +6,7 @@ namespace: argocd
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: apr
+namePrefix: progressive-sync-
```
sledigabel, line 9: the extra `-` in the end isn't gonna cause any issue i presume?

main.go
```diff
@@ -41,27 +46,31 @@ var (
...
 	flag.StringVar(&ctrlNamespace, "ctrl-namespace", "argocd", "The controller namespace")
 	flag.StringVar(&argoNamespace, "argo-namespace", "argocd", "The namespace where ArgoCD and the ApplicationSet controller are deployed to.")
-	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-
-	opts := zap.Options{}
+	opts := zap.Options{
+		Development: true,
```
sledigabel, line 69: that is going to generate tons of logs by default right? I'd rather we leave the line below add the dev flag on

main.go
```diff
@@ -99,7 +108,16 @@ func main() {
...
 		os.Exit(1)
 	}
-	// +kubebuilder:scaffold:builder
+	//+kubebuilder:scaffold:builder
+
+	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+		setupLog.Error(err, "unable to set up health check")
+		os.Exit(1)
+	}
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
```
sledigabel, line 117: this is a very interesting addition.
Worth adding an issue to specialise the ready check?
For instance, making sure it can read the argo config etc

sledigabel (approved):
LGTM, migrates the project to kubebuilder3.
Some minor comments, and also I'm assuming that most (all ) of the changes are pretty much boilerplate

Dockerfile
maruina: To be honest, this is something generated by kubebuilder. I can have a look why they do that, but I'm not 100% sure.

Makefile
maruina: This is coming from kubebuilder but I can add that

config/default/kustomization.yaml
maruina: No I think the old one is wrong but since we don't use kustomize we never realized that. This is a prefix so it makes sense to have the `-`

main.go
maruina: Yes! :)

main.go
maruina: I _think_ that what changes is the log format. Looking at https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap@v0.6.4#Options it says 
```
	// Development configures the logger to use a Zap development config
	// (stacktraces on warnings, no sampling), otherwise a Zap production
	// config will be used (stacktraces on errors, sampling).
```
Doesn't sound too bad but I can change it back if we want

Makefile
maruina: So those are our two targets:
- `test: manifests generate fmt shfmt vet ginkgo ## Run tests.`
- `build: generate fmt vet test ## Build manager binary.`
It seems to me that if we add test to build we have a lot of repetitions. What do you think?

main.go
sledigabel: I was under the impression it would activate all the `DEBUG` traces, i was prob wrong.
We can start like that and see what happens.

Issue #120, opened 2021-07-09
Stable/Deterministic Scheduling


nebojsa-prodana:
Right now as it stands, the current scheduler only returns us back the next applications to be scheduled during the iteration of the reconciliation loop. This makes the logic for checking whether the stage that we are currently working on is progressing/failing/completed significantly more difficult to assert as we have to observe the state of the world in each individual method.

Different runs of the reconciliation loop effectively lead to a different state of the observed world and handling all of these cases in each individual IsStage* method is becoming unwieldy. 

I propose that instead of that approach we test and pivot to an approach where the Scheduler is able to observe the complete state of the world and come up with an action plan to be executed per each stage. This would involve:
- Observing applications that are already healthy and synced and tying them up with the stage that synced them
- Observing applications that are progressing regardlessly of whether they are synced or not and plan them for a certain stage
-  Observing healthy out of sync applications, running them through stage selectors and planning them for the latter execution
-  Observing degraded applications and retrying them or failing the rollout
-  Observing the status of the clusters that are being selected per stage and planning out accordingly what to do with the deployment in case that the cluster went offline or are being traffic drained etc. Should we wait? Reschedule the cluster for a later stage or for a later run of the current stage based on maxTargets/maxParallel? TBD
- Being able to recover from failures: if we end up keeping the last observed state of the world in some storage (annotations/config map) the reconciler and the scheduler should still be able to recover in case of network errors or the controller pods crashing.

Issue #99, opened 2021-06-14
Remove predicate.GenerationChangedPredicate{}


maruina:
In https://github.com/Skyscanner/applicationset-progressive-sync/pull/96 we added a predicate to the `For` responsible for sending ProgressiveSync events.
While that helped us moving forward, it means the controller is not able to auto-heal in case something changes the CRD Status.
We should remove the predicate and make the reconciliation loop idempotent, so it should converge over time to the desired state.

maruina:
Fixed in #143

PR #143, opened 2021-11-07
New scheduler

maruina:
Fix https://github.com/Skyscanner/applicationset-progressive-sync/issues/120, https://github.com/Skyscanner/applicationset-progressive-sync/issues/99, https://github.com/Skyscanner/applicationset-progressive-sync/issues/115

I am aware that is not a great example of a PR, because it's pulling in too many changes at the same time, but I found much easier to do this refactor right now rather that using the old building blocks. Apologies to the reviewers.

### Notes
- Built on top of https://github.com/Skyscanner/applicationset-progressive-sync/pull/139
- Decided *against* adding a new `Stage` CR as stage objects, because:
  - it would have generated too many CR. Assuming an average 3 stages per deployment it would result in 3 new Kubernetes objects per ProgressiveSync.
  - The `stage.spec` would be a repetition of the `progressivesync.spec.stages[]` without adding any extra useful information
- We had an internal conversation and we agreed to take a step back, simplify the `status` API and focus on the reconciliation logic. Because of that:
  - Removed the `[]StageStatus` in favour to a simpler apporach by using `` and ``. This is going to be much simpler that handling the list update.
  - Removed the `sourceRef` in in favour to a simpler apporach by using `appSetRef` that points to an ApplicationSet in the controller namespace
  - Removed the multiple namespace controll to have consistency with `ArgoCD` and the `ApplicationSet` controller as they watch resources in a single namespace.
  - Dropped the support for `MaxTargets` and `MaxParallel` as %. At the moment we perform the calculation only when the controller starts so  it doesn't provide a lot of value. This allow us to have a smaller state object.
 - Shortened some variables names for the ProgressiveSync `spec`
 - Implement the `observedGeneration` pattern to tell for which `generation` the `status` is valid.
- The state is now stored in a ConfigMap (one per ProgressiveSync) instead of being in memory
- Used github.com/fluxcd/pkg/ as library for some common operations (condtions)
- Fix the hash function by using the `spew` library
- Fix printing the stacktrace when requing an object with error. See https://github.com/kubernetes-sigs/kubebuilder/issues/1593
- Switch from `ginkgo` to `go test` for Integration tests, following the ClusterAPI testing strategy (see https://cluster-api.sigs.k8s.io/developer/testing.html and https://cluster-api.sigs.k8s.io/developer/testing.html#ginkgo) and https://github.com/fluxcd/kustomize-controller/blob/main/controllers/suite_test.go. We will use `gingko` for writing real e2e tests once we have scripts to create multiple local clusters). This makes the tests faster and easier to read.


controllers/progressivesync_controller.go
```diff
@@ -341,168 +281,272 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
...
+		// If we don't have a state for a synced app
+		// it means it was synced by an external process.
+		// Adopt the app by setting its synced stage to the current one.
+		if !ok {
+			state.Apps[app.Name] = AppState{
+				SyncedAtStage: stage.Name,
 			}
-			log.Info("failed to sync app because it is already syncing")
+			syncedInCurrentStage = append(syncedInCurrentStage, app)
+			break
```
Smirl, line 429:
```suggestion
			continue
```

Smirl: could also be an `else if`

controllers/progressivesync_controller.go
```diff
@@ -341,168 +281,272 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
...
-		return ps, ctrl.Result{Requeue: true, RequeueAfter: RequeueDelayOnError}, nil
+	// If there are no out-of-sync apps then the stage is completed
+	if len(outOfSyncApps) == 0 {
+		return syncv1alpha1.StageStatusCompleted, nil
 	}
 
-	if scheduler.IsStageInProgress(apps, stage, pss) {
-		message := fmt.Sprintf("%s stage in progress", stage.Name)
-		log.Info(message)
+	log.Info("apps summary", "progressingApps", utils.GetAppsName(progressingApps), "syncedInCurrentStage", utils.GetAppsName(syncedInCurrentStage), "OutOfSync", utils.GetAppsName(outOfSyncApps), "stage", stage.Name)
```
Smirl, line 469:
```suggestion
```

controllers/progressivesync_controller.go
```diff
@@ -341,168 +281,272 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
...
+		maxSync = len(outOfSyncApps)
+	}
 
-		r.updateStageStatus(ctx, stage.Name, message, syncv1alpha1.PhaseProgressing, &ps)
+	// If the number of apps synced in the current stage
+	// is equal or greater than the maximum number of targets to sync,
+	// there is nothing else to do.
+	if len(syncedInCurrentStage) >= maxTargets {
+		return syncv1alpha1.StageStatusCompleted, nil
+	}
```
Smirl, line 484: perhaps move both completed checks next to each other

controllers/progressivesync_controller.go
```diff
@@ -341,168 +281,272 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
...
-		log.Info(message)
+	log.Info("apps summary", "progressingApps", utils.GetAppsName(progressingApps), "syncedInCurrentStage", utils.GetAppsName(syncedInCurrentStage), "OutOfSync", utils.GetAppsName(outOfSyncApps), "stage", stage.Name)
 
-		for _, app := range apps {
-			log.Info("application details", "app", fmt.Sprintf("%s/%s", app.Namespace, app.Name), "sync.status", app.Status.Sync.Status, "health.status", app.Status.Health.Status)
-		}
+	// If there is an external process triggering a sync,
+	// maxParallel - len(progressingApps) might actually be greater than len(outOfSyncApps)
+	// causing the runtime to panic
+	maxSync := maxParallel - len(progressingApps)
```
Smirl, line 474: rather than creating a new variable, could we update the value of `maxParallel` if we don't need it again?

Smirl (changes requested):

api/v1alpha1/progressivesync_types.go
```diff
@@ -17,90 +17,132 @@ limitations under the License.
...
+type Stage struct {
 	// Name is a human friendly name for the stage
 	//+kubebuilder:validation:Required
 	Name string `json:"name"`
-	// MaxParallel is how many selected targets to update in parallel
+
+	// MaxParallel is how many targets to sync in parallel
+	//+kubebuilder:validation:Required
 	//+kubebuilder:validation:Minimum:1
-	MaxParallel intstr.IntOrString `json:"maxParallel"`
```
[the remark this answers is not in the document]
Smirl: https://github.com/Skyscanner/applicationset-progressive-sync/issues/112

charts/applicationset-progressive-sync/crds/argoproj.skyscanner.net_progressivesyncs.yaml
```diff
@@ -0,0 +1,237 @@
...
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: ProgressiveSyncSpec defines the desired state of ProgressiveSync
+            properties:
+              sourceRef:
```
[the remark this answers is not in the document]
maruina: Yes, I'm going to raise a separate PR where I update the chart because otherwise the chart test breaks. 
This is because `ct` tries to pull an actual image.

controllers/progressivesync_controller.go
```diff
@@ -341,168 +281,272 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
...
+	}
+
+	// Create the progressive sync state configmap if it doesn't exist
+	if err := r.CreateStateMap(ctx, ps); err != nil {
+		log.Error(err, "unable to create state map")
+		return ps, ctrl.Result{RequeueAfter: RequeueDelayOnError}, err
+	}
+
+	// Read the latest state
+	stateData, err := r.ReadStateMap(ctx, ps)
```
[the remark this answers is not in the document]
maruina: Sure. I named it `state` everywhere

controllers/progressivesync_controller.go
maruina: Yep, done.

controllers/progressivesync_controller.go
maruina: Done

Smirl (approved):
Great PR which cleans up the logic, and simplifies the controller. Good job!

README.md
```diff
@@ -73,16 +72,15 @@ The above configuration is loaded taking into account the following priority ord
...
 
-
-```
+```console
 ARGOCD_AUTH_TOKEN=ey...
 ARGOCD_SERVER_ADDR=argocd-server
 ARGOCD_INSECURE=true
 ```
 
+1. Files in the Config Directory (`/etc/prc-config/`).
```
sledigabel, line 81: small thing, but `prc-config` doesn't line up with the `prcconfig` below.
And should we rename this `psc-config`?

sledigabel (commented):
Some minor comments, I still need to review the main one, `progressive_controller.go`.

api/v1alpha1/progressivesync_types.go
```diff
@@ -17,90 +17,132 @@ limitations under the License.
...
-	// MaxTargets is the maximum number of selected targets to update
+	MaxParallel int64 `json:"maxParallel"`
```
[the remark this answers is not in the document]
maruina: Because I thought the official guidance was to use int64 but it's that or int32 (https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types)

I can convert it to int32

controllers/progressivesync_controller.go
```diff
@@ -341,168 +281,265 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
...
+	if ps.Status.ObservedGeneration != ps.Generation {
+		ps.Status.ObservedGeneration = ps.Generation
+		return syncv1alpha1.ProgressiveSyncProgressing(ps), ctrl.Result{Requeue: true}, nil
+	}
 
-// reconcileStage reconcile a ProgressiveSyncStage
-func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv1alpha1.ProgressiveSync, stage syncv1alpha1.ProgressiveSyncStage, pss utils.ProgressiveSyncState) (syncv1alpha1.ProgressiveSync, reconcile.Result, error) {
-	log := r.Log.WithValues("progressivesync", fmt.Sprintf("%s/%s", ps.Namespace, ps.Name), "applicationset", ps.Spec.SourceRef.Name, "stage", stage.Name, "syncedAtStage", pss)
+	// Create the progressive sync state configmap if it doesn't exist
+	if err := r.CreateStateMap(ctx, ps); err != nil {
```
sledigabel, line 310: I'm not sure how to rename the function, but it seems to return no error if the statemap exists already (which makes sense, but it doesn't _create_ it)

controllers/progressivesync_controller.go
```diff
@@ -341,168 +281,265 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
...
+		log.Error(err, "unable to create state map")
+		return ps, ctrl.Result{RequeueAfter: RequeueDelayOnError}, err
+	}
 
-	// Get the clusters to update
-	clusters, err := r.getClustersFromSelector(ctx, stage.Targets.Clusters.Selector)
+	// Read the latest state
+	state, err := r.ReadStateMap(ctx, ps)
```
sledigabel, line 316: why not read directly and create the statemap if it doesn't exist directly?

controllers/progressivesync_controller.go
```diff
@@ -341,168 +281,265 @@ func (r *ProgressiveSyncReconciler) getOwnedAppsFromClusters(ctx context.Context
...
+	//
+	// maxTargets = 3
+	// maxParallel = 3
+	// outOfSyncApps = 4
+	// syncedInCurrentStage = 2
+	// progressingApps = 1
+	//
+	// Without the following logic we would end up
+	// with a total of 4 applications synced in the stage
+	if maxParallel+len(syncedInCurrentStage) > maxTargets {
```
sledigabel, line 491: the logic seems exactly the same as below, but it's a bit confusing the way it's presented.
I think we should refactor it with the lines above.
```
currentParallel := min(maxParallel - len(progressingApps), len(outOfSync), maxTargets - len(syncedInCurrentStage)
```

I think it would make the logic clearer.

sledigabel (commented):
Sorry it took so long to review.
This is great work, I much prefer it the way it is now.

A few small comments

api/v1alpha1/progressivesync_types.go
```diff
@@ -17,90 +17,132 @@ limitations under the License.
...
 	//+kubebuilder:validation:Optional
-	Targets ProgressiveSyncTargets `json:"targets,omitempty"`
+	Targets Targets `json:"targets,omitempty"`
 }
 
-// ProgressiveSyncTargets defines the target of the Progressive Rollout
-type ProgressiveSyncTargets struct {
-	// Clusters is the a cluster type of targets
+// Targets defines the targets of the progressive sync operation
+type Targets struct {
```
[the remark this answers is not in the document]
maruina: Good question. I guess yes if we wanted to introduce the ability to target Applications

controllers/progressivesync_controller.go
```diff
@@ -57,106 +51,82 @@ const RequeueDelayOnError = time.Minute * 5
...
-		if err := r.updateStatusWithRetry(ctx, &latest); err != nil {
-			return ctrl.Result{}, err
-		}
+	mapOwnerKey := ".metadata.controller"
+	apiGVStr := syncv1alpha1.GroupVersion.String()
 
-		if result.Requeue || reconcileErr != nil {
-			log.Info("requeuing stage")
-			return result, reconcileErr
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.ConfigMap{}, mapOwnerKey, func(rawObj client.Object) []string {
```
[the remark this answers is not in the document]
maruina: It's adding an index to allow our reconciler to quickly look up configmaps by their owner. See https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#setup

controllers/progressivesync_controller.go
maruina: I think the idea here was to have a CRUD approach, so this is why create, read, update and delete

controllers/progressivesync_controller.go
maruina: I think because this is how it came along :)

At the moment we need to always call create before reading, otherwise we might want to try to read something that doesn't exist. I can change the logic here

controllers/progressivesync_controller.go
maruina: Before the code review with @Smirl I was using a variable called `maxSync` and the code was
```
	maxSync := maxParallel - len(progressingApps)
	if maxSync > len(outOfSyncApps) {
		maxSync = len(outOfSyncApps)
	}
```
Since we were using maxParallel only to calculate maxSync, @Smirl suggested to overwrite maxParallel to avoid adding a new variable.

I'm happy with both approach to be honest, but I would probably like to keep the calculation explicit rather than a 1 liner because I found it easier to understand.

@Smirl what do you think?

controllers/progressivesync_state.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	key := getStateMapNamespacedName(ps)
+
+	cm := corev1.ConfigMap{}
+	if err := r.Get(ctx, key, &cm); err != nil {
+		return err
+	}
+
+	cm.Data = map[string]string{
```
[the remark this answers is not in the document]
maruina: because cm.Data is a `map[string]string` and you can't directly unmarshal into the state object. @Smirl tried it and the compiler was complaining.

api/v1alpha1/progressivesync_types.go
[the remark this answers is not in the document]
maruina: Discussed offline and decided to keep `int64`

controllers/progressivesync_controller.go
maruina: Done

controllers/progressivesync_controller.go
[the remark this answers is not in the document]
sledigabel: worth adding this link as a comment for future refs?

controllers/progressivesync_controller.go
sledigabel: 💯  on the explicit, but the current calculation doesn't make it obvious that you're basically trying to calculate a low water mark through 3 values and it took me a while to read through it in views with lines 470-473

controllers/progressivesync_controller.go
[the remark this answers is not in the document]
maruina: Done

controllers/progressivesync_controller.go
maruina: So, should we go back to the previous code with `maxSync`?

README.md
maruina: Fixed with a more explicit `applicationset-progressive-sync`

controllers/progressivesync_controller.go
maruina: Pushed

PR #148, opened 2021-12-13
allow to configure client plaintext connection

riuvshyn:
hello! thanks for very useful argo extention!
Please consider following change:
allow to configure client plaintext connection, when applicationset-progressive-sync is running in the same cluster with argocd this will allow to use argocd svc as address `ARGOCD_SERVER_ADDR=argocd-server:80`


maruina:
Hey @riuvshyn, thanks for the contribution. I will fix the build process in the next couple of days so we will see all green on this PR and then we can merge it.

maruina:
Hey @riuvshyn , once https://github.com/Skyscanner/applicationset-progressive-sync/pull/149 it's merged you should be able to:

- rebase
- bump the helm chart to fix the chart test failing

and then we can merge this

maruina (approved):

PR #154, opened 2022-01-18
Update helm chart and fix docker build

maruina:
- Update the tags when building a new docker image. At the moment we just generate the short sha, for example `sha-f70b92b`.  This PR enable generating a `latest`, `<branch>` and a `semver` tags when cutting a release. See https://github.com/docker/metadata-action
- Update the Helm chart to pull the latest controller version
- Drop support for HELM v2 since it's EOL since November 2020, see [https://helm.sh/blog/helm-v2-deprecation-timeline/](https://helm.sh/blog/helm-v2-deprecation-timeline/)
- Fix the arguments passed to the manager with the changes introduced in #143 
- Switch from installing only the ArgoCD CRD to the whole ArgoCD, otherwise the controller complains that can't talk to argocd server
- Fix leader election installation and add a test for it
- Fix argocd servic address in existing secret

Unfortunately `ct` cares only if a pod can start correctly, but if the controller it's throwing errors you're going to miss it. I ran the tests locally and checked the output of every test case to fix the issues here. I'm not 100% sure what are the best practices for that.


sledigabel (approved):
LGTM, updates the helm chart versions, updates the CI, and publishes new tags for the docker image.

Issue #147, opened 2021-12-09
Sometimes stages executed in parallel


riuvshyn:
sometimes stages executed in parallel potentially due to some race condition

I've noticed this unexpected behaviour quite a couple times already when apps are going outOfSync due to git change and getting executed in parallel without staging. 

maybe this happens because status is not instantly getting updated due to k8s eventual consistency or in place where app sync is invoked or so...

ProgressiveSync ex:
```
apiVersion: argoproj.skyscanner.net/v1alpha1
kind: ProgressiveSync
metadata:
  labels:
    argocd.argoproj.io/instance: argocd-apps
  name: test
  namespace: argocd
spec:
  sourceRef:
    apiGroup: argoproj.io/v1alpha1
    kind: ApplicationSet
    name: test
  stages:
  - maxParallel: 1
    maxTargets: 100%
    name: finish rollout staging
    targets:
      clusters:
        selector:
          matchLabels:
            environment: staging
  - maxParallel: 1
    maxTargets: 100%
    name: finish rollout production
    targets:
      clusters:
        selector:
          matchLabels:
            environment: production
```

riuvshyn:
this can be reproduced reliably with github webhook so a change in a repo almost immediately changes state to outOfSync for many applications at once. 

If we add a sleep for few seconds in here https://github.com/Skyscanner/applicationset-progressive-sync/blob/main/controllers/progressivesync_controller.go#L137 problem goes away so it looks indeed like race condition

maruina:
I'd like to add a test for this in https://github.com/Skyscanner/applicationset-progressive-sync/pull/143, do you have any suggestion?

maruina:
Thank you very much for your report.

> expected behaviour was: canary stg -> finish stg -> canary prod -> finish prod but in fact stages were executed in parallel in random order.

This is interesting. This might happen because the controller is missing an event and then it works on different assumption. I think the only way to reproduce this is with an E2E test. In the integration tests we are the one chaning the Applications status, so it's going to be tricky to reproduce.

How long do you sleep for in your patch? I wonder if it would make sense to add the sleep in upstream until we can reproduce this with a test. Withour a repro is going to be difficult to fix it

> So it might turn out that production state was updated to outOfSync first then progressive-sync kicks-in and it see that staging is synced (because it is not yet evaluated I think) and progressive-sync starts syncing production straight away while expected behaviour is stg should be always synced first and then production...

[...]

PR #157, opened 2022-01-28
Ensure all the Applications have the same revision

maruina:
Fix an issue reported in https://github.com/Skyscanner/applicationset-progressive-sync/issues/147#issuecomment-1020074054

When the remote repository is updated, ArgoCD starts marking all the Applications pointing to it as `out of sync`.

This PR ensures the controller doesn't start until ArgoCD has updated all the Applications.


sledigabel (approved):
LGTM. Validates that apps have the same revision before proceeding to the progressive sync.

Smirl (approved):

Issue #146, opened 2021-12-09
stage is stuck forever if targeted 0 clusters


riuvshyn:
when there are 4 stages
- stg canary
- stg other clusters
- prod canary
- prod other clusters

but there are only `1` cluster for stg and `1` for prod second stage is going into `Progressing` phase and stuck there forever.

I guess expected behaviour would be skip stage if it did not found other clusters or at least have this as an optional flag. 

ProgressiveSync object ex:
```
apiVersion: argoproj.skyscanner.net/v1alpha1
kind: ProgressiveSync
metadata:
  labels:
    argocd.argoproj.io/instance: argocd-apps
  name: test
  namespace: argocd
spec:
  sourceRef:
    apiGroup: argoproj.io/v1alpha1
    kind: ApplicationSet
    name: test
  stages:
  - maxParallel: 1
    maxTargets: 1
    name: canary rollout staging
    targets:
      clusters:
        selector:
          matchLabels:
            environment: staging
  - maxParallel: 1
    maxTargets: 100%
    name: finish rollout staging
    targets:
      clusters:
        selector:
          matchLabels:
            environment: staging
  - maxParallel: 1
    maxTargets: 1
    name: canary rollout production
    targets:
      clusters:
        selector:
          matchLabels:
            environment: production
  - maxParallel: 1
    maxTargets: 100%
    name: finish rollout production
    targets:
      clusters:
        selector:
          matchLabels:
            environment: production
status:
  conditions:
  - lastTransitionTime: "2021-12-09T10:50:24Z"
    message: finish rollout staging stage in progress
    reason: StagesProgressing
    status: "False"
    type: Completed
  stages:
  - finishedAt: "2021-12-09T14:34:42Z"
    message: canary rollout staging stage completed
    name: canary rollout staging
    phase: Succeeded
  - message: finish rollout staging stage in progress
    name: finish rollout staging
    phase: Progressing
    startedAt: "2021-12-09T14:34:42Z"
  - finishedAt: "2021-12-09T12:16:43Z"
    message: canary rollout production stage completed
    name: canary rollout production
    phase: Succeeded
  - finishedAt: "2021-12-09T12:13:50Z"
    message: finish rollout production stage completed
    name: finish rollout production
    phase: Succeeded
```

maruina:
[...]

PR #159, opened 2022-02-03
skip stage if no clusters were targeted

riuvshyn:
fix/workaround for https://github.com/Skyscanner/applicationset-progressive-sync/issues/146

basically the idea is to skip stage if targets returned empty list.

This also deserves a test but atm I do not have capacity for this :(

### DIFF
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -372,7 +372,6 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
 
 	// Skip stage if no clusters targeted
 	if len(selectedClusters.Items) == 0 {
-		log.Info("Skip stage as no clusters targeted", "stage", stage.Name)
 		return syncv1alpha1.StageStatusCompleted, nil
 	}
 
--- a/controllers/progressivesync_controller.go
+++ b/controllers/progressivesync_controller.go
@@ -369,6 +369,13 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
 	if err != nil {
 		return syncv1alpha1.StageStatusFailed, err
 	}
+
+	// Skip stage if no clusters targeted
+	if len(selectedClusters.Items) == 0 {
+		log.Info("Skip stage as no clusters targeted", "stage", stage.Name)
+		return syncv1alpha1.StageStatusSkipped, nil
+	}
+
 	// Get the ArgoCD apps targeting the selected clusters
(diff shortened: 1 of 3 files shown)


maruina:
Hey, @riuvshyn thanks for the PR.

I had a similar thought about solving this issue as well, glad to see I'm not alone :)

A couple of things:
- I'm not sure about adding a new `StageStatus`. On one side I can see how having a status to tell you "hey, you're targeting 0 clusters, are you sure?" is a little more indicative. On the other side, it feels more like a `Completed` stage to me. If you select 0 clusters with the label selector, that is a valid choice. Can you tell me why you went for a new `StageStatus`?
- The test would be great, but if you don't have the capacity don't worry e I'll give it a try.

riuvshyn:
heya @maruina, 
re `Skipped` status I thought it might be useful in future but I agree `Completed` is absolutely fine.
So i've used `Completed` status.
Also I've noticed that test was there already but it wasn't doing "missing" cluster  stage, so I've updated test and it should cover now this case.

controllers/progressivesync_controller.go
```diff
@@ -369,6 +369,13 @@ func (r *ProgressiveSyncReconciler) reconcileStage(ctx context.Context, ps syncv
 	if err != nil {
 		return syncv1alpha1.StageStatusFailed, err
 	}
+
+	// Skip stage if no clusters targeted
+	if len(selectedClusters.Items) == 0 {
+		log.Info("Skip stage as no clusters targeted", "stage", stage.Name)
```
maruina, line 375: Just a little thing. Can you remove this logline here?

I'd like to file a new issue to change the `reconcileStage` function to return a `reason` so we can have a clear understanding of what's happening, but I'd prefer to avoid the single logline at this time.

maruina (changes requested):

controllers/progressivesync_controller_test.go
```diff
@@ -365,6 +365,10 @@ func TestReconcile(t *testing.T) {
 
 		ps := newProgressiveSync("zero-clusters", ns.Name, appSet)
 		ps.Spec.Stages = []syncv1alpha1.Stage{
+			newStage("stage-missing-cluster", 1, 1, metav1.LabelSelector{
```
maruina, line 368:
```suggestion
			newStage("stage missing cluster", 1, 1, metav1.LabelSelector{
```
just to be consistent with the other stages

controllers/progressivesync_controller.go
riuvshyn: done

riuvshyn:
ty for the review @maruina, all done

maruina (approved):
Thanks for the PR, LGTM

Issue #129, opened 2021-07-28
Allow pruning when syncing


maruina:
In same cases we need to be able to prune objects. The ArgoCD UI is pretty clear about this
<img width="674" alt="image" src="https://user-images.githubusercontent.com/4821896/127345284-a8b1bba6-a944-405f-8387-451042930357.png">

We need to either enable pruning by default or add a flag to enable it.

maruina:
Nope because the AppSet just updates the Application, but here the problem is what's inside the Helm chart we apply in the target cluster. When switching from a non-flagger to a flagger enabled service there were some stale resources around.

PR #172, opened 2022-04-07
adds sync options for the stage

riuvshyn:
hello, this is an attempt to fix https://github.com/Skyscanner/applicationset-progressive-sync/issues/129
currently the only sync option is `prune` but can be extended if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ spec:
   appSetRef:
     name: myappset
     # the rollout steps
+  # app sync options
+  syncOptions:
+    # enable prune resources during sync, disabled by default
+    prune: true
   stages:
       # human friendly name
     - name: two clusters as canary in EMEA
@@ -42,10 +46,6 @@ spec:
           selector:
             matchLabels:
               area: emea
-      # app sync options
-      syncOptions:
-        # enable prune resources during sync, disabled by default
-        prune: true
     - name: rollout to remaining clusters
       maxParallel: 2
       maxTargets: 4


riuvshyn:
@maruina heya, hope you are doing good! we would like to enable pruning with progressive sync controller so wondering WDYT about this?

maruina:
Hey @riuvshyn, sorry for the late reply, I was off for a few days. I think it's a good idea and I will have a look at the PR in the next couple of days

api/v1alpha1/progressivesync_types.go
```diff
@@ -57,6 +57,17 @@ type Stage struct {
 	// Targets is the targets to sync in the stage
 	//+kubebuilder:validation:Optional
 	Targets Targets `json:"targets,omitempty"`
+
+	// SyncOptions is the options for sync operation executed for the stage
+	//+kubebuilder:validation:Optional
+	SyncOptions SyncOptions `json:"syncOptions,omitempty"`
```
maruina, line 63: Do we want to keep this at the stage level? I'm asking because it feels a little heavy, because normally one should prune or not prune.

maruina (commented):
Apologies for the late review. I added a comment because having this at the stage level feels a little too granular. I'm interested in the rationale behind this approach

api/v1alpha1/progressivesync_types.go
riuvshyn: Was thinking that it might bring some more flexibility like you might want to enable pruning only in staging stages but not in production for some reason... I am completely fine to move this to `type ProgressiveSyncSpec struct` 
I don't have strong opinion on it, WDYT?

maruina: I would go for a more generic option to start lean, and then we can expand it if we find a valid use case.

riuvshyn: @maruina hey, sorry for delay, I've just moved syncOpts to the PS level so it will be applied for all stages.

maruina (approved):

